PR #3, opened 2022-09-04
Use Firebase authentication to identify users.

bharathcs:
Firebase will fingerprint the browser and assign an anonymous, stable
user id despite having no log-in user flow.

User can wipe existing sign in data (lose access to anonymous userId,
sign out of Github if applicable) with the Log Out button.

User can then replace this with a Firebase UID generated by their Github
account by clicking Sign in with Github.

See [discussion](https://docs.google.com/document/d/1cxMVBbcbmjTe-KS6NQpHviEETmU1qwFvYl0IIoJ2_K0/edit)
for more details on the decision and the steps forward.

:warning: `.env` file is needed, check firebase console for keys.


context/AuthContext.ts
```diff
@@ -0,0 +1,16 @@
+import { User, Auth } from 'firebase/auth';
+import { createContext } from 'react';
+
+type ContextProps = {
```
sivayogasubramanian:
```suggestion
type AuthContextProps = {
```

context/AuthContext.ts
```diff
@@ -0,0 +1,16 @@
...
+  currentUser: User | undefined;
+  setCurrentUser: (newUser: User) => void;
+  auth: Auth | undefined;
+};
+
+const AuthContext = createContext<ContextProps>({
```
sivayogasubramanian:
```suggestion
const AuthContext = createContext<AuthContextProps>({
```

sivayogasubramanian (approved):
LGTM! Didn't check the code for `index.tsx` as it will be changed later. But functionality is good!

PR #8, opened 2022-09-06
Create placeholder API files.

bharathcs:
- Add placeholder API files
- Add placeholder API logic.


zognin:
Thanks for setting up, siva tested the current endpoints and they work

- On hindsight siva and I think we don't need `/user/` before `/application/*` because it is unnecessary, shall we remove it?
  - If we decide to keep it then `/user/` can be changed to `/users/`
- Is `/companies` missing?

zognin (commented):
Thanks for setting up, siva tested the current endpoints and they work

- On hindsight siva and I think we don't need `/user/` before `/applications/*` because it is unnecessary, shall we remove it?
  - If we decide to keep it then `/user/` can be changed to `/users/`
- Is `/companies` missing?

bharathcs:
Sgtm, thanks for the catch too.

sivayogasubramanian (approved):
LGTM!

PR #9, opened 2022-09-06
Add Prisma models and seed data

ganhongyao:
### Changes
- Set up Amazon RDS databases `caza` and `caza-preview` 
- Add Prisma models
- Add seed data

### To get your databases created, migrated and seeded:

1. Install new dependencies
```bash
yarn
```
2. Make a copy of `.env.example` as `.env`.
```bash
cp .env.example .env
```
3. Fill in the `DATABASE_URL` in `.env`, replacing `USER` and `PASSWORD` with the actual values.
4. Run 
```bash
yarn prisma migrate dev
```


prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+  updatedAt DateTime @updatedAt @map("updated_at")
+
+  @@map("application_stages")
+}
+
+model Reaction {
+  id               Int               @id @default(autoincrement())
+  emojiUnicodeHex  String            @map("emoji_unicode_hex")
+  remark           String
+  applicationStage ApplicationStage?
```
bharathcs: Another minor point: if the mapping is `ApplicationStage (1) <-> (0 or 1) Reaction` shouldn't the applicationStage not be nullable or is this a limitation of how Prisma models work?

.env.example
```diff
@@ -5,3 +5,6 @@ NEXT_PUBLIC_FIREBASE_PROJECT_ID=""
 NEXT_PUBLIC_FIREBASE_STORAGE_BUCKET=""
 NEXT_PUBLIC_FIREBASE_MESSAGING_SENDER_ID=""
 NEXT_PUBLIC_FIREBASE_APP_ID=""
+
+# Database
```
bharathcs: Super nit to differentiate btwn firebase and pg

```suggestion
# PostgreSQL Database
```

bharathcs (approved):
Nice work! I've also added the env vars (from telegram) to the [env doc](https://docs.google.com/document/d/1LNdXt-Ad7vx9XJE1JJOPV1vVJUldhV1PPfK2UcTglyo/edit).

prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+  updatedAt DateTime @updatedAt @map("updated_at")
+
+  applications Application[]
+
+  @@map("roles")
+}
+
+model Application {
+  id     Int    @id @default(autoincrement())
+  user   User   @relation(fields: [userId], references: [uid])
```
sivayogasubramanian: Should we change to ON DELETE CASCADE for this?

prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+  userId String @map("user_id")
+  role   Role   @relation(fields: [roleId], references: [id])
```
sivayogasubramanian: Should we change to ON DELETE CASCADE for this?

prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+  WINTER_INTERNSHIP
+  SPRING_INTERNSHIP
+  SUMMER_INTERNSHIP
+  FALL_INTERNSHIP
+  FULL_TIME
+}
+
+model Role {
+  id         Int      @id @default(autoincrement())
+  company    Company  @relation(fields: [companyId], references: [id])
```
sivayogasubramanian: Should we change to ON DELETE CASCADE for this?

prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+  companyId  Int      @map("company_id")
+  title      String
+  type       RoleType
+  year       Int
```
sivayogasubramanian, line 51: Should we add a CHECK constraint here to check if this is really a valid year (current or future)? Not sure how it is done in Prisma.

prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+  MIXED
+  OFFERED
+  ACCEPTED
+  REJECTED
+  WITHDRAWN
+}
+
+model ApplicationStage {
+  id            Int                  @id @default(autoincrement())
+  application   Application          @relation(fields: [applicationId], references: [id])
```
sivayogasubramanian: Should we change to ON DELETE CASCADE for this?

prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+
+  @@map("application_stages")
+}
+
+model Reaction {
+  id               Int               @id @default(autoincrement())
+  emojiUnicodeHex  String            @map("emoji_unicode_hex")
+  remark           String
+  applicationStage ApplicationStage?
+
```
sivayogasubramanian:
```suggestion
  applicationStage ApplicationStage? @relation(fields: [applicationStageId], references: [id])
  applicationStageId Int? @map("application_stage_id")
```

sivayogasubramanian: Also, should we do ON DELETE CASCADE for this?

prisma/schema.prisma
```diff
@@ -0,0 +1,126 @@
...
+  updatedAt DateTime @updatedAt @map("updated_at")
+
+  @@map("reactions")
+}
+
+model Task {
+  id                   Int         @id @default(autoincrement())
+  application          Application @relation(fields: [applicationId], references: [id])
```
sivayogasubramanian: Should we do ON DELETE CASCADE for this?

prisma/migrations/20220906193601_add_initial_models/migration.sql
```diff
@@ -0,0 +1,106 @@
...
+    "updated_at" TIMESTAMP(3) NOT NULL,
+
+    CONSTRAINT "application_stages_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "reactions" (
+    "id" SERIAL NOT NULL,
+    "emoji_unicode_hex" TEXT NOT NULL,
+    "remark" TEXT NOT NULL,
```
zognin, line 65: Must users always add a remark? Is it possible to put an emoji without a remark?

prisma/seed.ts
```diff
@@ -0,0 +1,196 @@
...
+
+main()
+  .then(async () => {
+    await prisma.$disconnect();
+  })
+  .catch(async (e) => {
+    console.error(e);
+    await prisma.$disconnect();
+    process.exit(1);
+  });
```
sivayogasubramanian, line 196: .finally( await prisma.$disconnect(); )

prisma/schema.prisma
ganhongyao: Yup, this is a restriction imposed by Prisma. Since we do not define the `applicationStageId` on the `Reaction` model, it has to optional.

See: https://www.prisma.io/docs/concepts/components/prisma-schema/relations/one-to-one-relations#required-and-optional-1-1-relation-fields

prisma/schema.prisma
bharathcs: I don't believe we should allow roles to be deleted if applications exist. I don't think we should allow delete from Role (admin access) if it affects apps (user data). Admin should manually delete or handle the applications to protect user data before deleting the role.

prisma/schema.prisma
ganhongyao: Yup, added cascade on updates and deletes for this and the below comments

prisma/schema.prisma
ganhongyao: Yup, added cascade on updates and deletes

prisma/migrations/20220906193601_add_initial_models/migration.sql
bharathcs: I would prefer forcing empty string over null for text and I can see how some users want to write text without emojis.

prisma/schema.prisma
ganhongyao: We already have `reaction_id` defined on the `application_stages` table, so from the SQL perspective we should not define cyclical foreign keys on both tables. Prisma enforces this as well. We can however choose which side holds the scalar i.e. `xId` ([Prisma docs](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/one-to-one-relations#choosing-which-side-should-store-the-foreign-key-in-a-1-1-relation)), and I think we chose to store `reaction_id` in `application_stages` from our schema discussion yesterday.

prisma/schema.prisma
ganhongyao: Do we really want this constraint though? I see potential use cases for allowing people to add applications for previous cycles.

prisma/migrations/20220906193601_add_initial_models/migration.sql
ganhongyao: I thought that the empty string would be fine, but on second thought perhaps `null` would be better.

Changed to make it optional.

ganhongyao: Oops didn't see @bharathcs 's comment when I wrote the reply, but I am open for discussion.

zognin: I think if either can be null, and there is at most 1 reaction per stage, then we can just put these columns into application_stages and do without a reactions table.

Otherwise if we keep reactions table, we put a check to ensure at least 1 of the fields is not null, but that is an extra step. Makes it harder if people remove their reactions after adding them -- do we delete or soft delete?

bharathcs: On nullity: My thinking was that we will not have different behaviour for `null` vs `''`, but I'm not opposed to allowing nulls, it only means there's an additional check in the client side.

On reactions: I agree with removing the reactions table and adding nullable columns. The downside is that its less extensible if we want to expand to do more than just a simple text box (e.g. additional records / images etc). I think that's a very minor reason though, and to avoid complexity we should just remove the table and simplify the design.

prisma/schema.prisma
sivayogasubramanian: Okay I think we can leave it as it is for now.

prisma/migrations/20220906193601_add_initial_models/migration.sql
ganhongyao: Removed reactions table and moved fields into `application_stages` table in commit https://github.com/sivayogasubramanian/caza/pull/9/commits/1660732a9d67ada754e80d624bb62d215c162237

prisma/schema.prisma
ganhongyao: I personally think it's fine to have `ON DELETE CASCADE`. `ON DELETE CASCADE` just makes it more dangerous if we actually do delete roles, but then again this is a rare case and I think for now we can assume that the admin takes active steps to protect user data before running the delete command.

PR #10, opened 2022-09-07
Add JWT middleware and testing utilities.

bharathcs:
Adds JWT middleware that can be used to simplify writing handlers. See `api/user/hello*` for examples. These will be removed once everyone is familiar with JWT.

If you are using postman / curl / etc, remember to set your Authorization header to Bearer type with the JWT token (must be stripped of trailing whitespace). Note that the token created by Firebase have a lifespan of 1 hour.

- Add JWT middleware.
- Add user/hello* endpoints to test calls with JWT.
- Add UI component to see JWT for testing purposes.

### DIFF
--- a/utils/auth/jwtHelpers.ts
+++ b/utils/auth/jwtHelpers.ts
@@ -63,7 +63,7 @@ export function withVerifiedUser<T>(
 
     const { uid, isAnonymous } = userDetails;
     if (isAnonymous) {
-      return res.status(401).end(ANONYMOUS_USER_NOT_AUTHORIZED);
+      return res.status(403).end(ANONYMOUS_USER_NOT_AUTHORIZED);
     }
 
     return handler(uid, req, res);
--- a/utils/auth/jwtHelpers.ts
+++ b/utils/auth/jwtHelpers.ts
@@ -20,7 +20,7 @@ try {
 }
 
 const UNABLE_TO_AUTHENTICATE = 'User could not be authenticated.';
-const ANONYMOUS_USER_NOT_AUTHORIZED = 'User has not verified their identity.';
+const USER_NOT_AUTHORIZED = 'User has not verified their identity.';
 
 /**
  * Utility function that takes care of authenticating any user (anonymous / verified) and providing the UID.
--- a/utils/auth/jwtHelpers.ts
+++ b/utils/auth/jwtHelpers.ts
@@ -1,0 +1,105 @@
+import admin from 'firebase-admin';
+import { NextApiHandler, NextApiRequest, NextApiResponse } from 'next';
+
+// Note: do not modify the 'firebase-admin' default import style, it triggers a known bug
+// See: https://github.com/firebase/firebase-admin-node/issues/593
+try {
+  admin.initializeApp({
+    projectId: process.env.NEXT_PUBLIC_FIREBASE_PROJECT_ID,
+  });
+} catch (error) {
+  // Ignore the "already exists" message in dev environments (cause by hot reloads).
+  if (!(process.env.NODE_ENV == 'development' && /already exists/u.test('' + error))) {
+    console.error(error);
+  }
+}
+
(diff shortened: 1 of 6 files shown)


utils/auth/jwtHelpers.ts
```diff
@@ -0,0 +1,105 @@
...
+    let userDetails: UserDetailsFromRequest;
+    try {
+      userDetails = await getUserFromJwt(getBearerToken(req));
+    } catch (e) {
+      return res.status(401).end(UNABLE_TO_AUTHENTICATE);
+    }
+
+    const { uid, isAnonymous } = userDetails;
+    if (isAnonymous) {
+      return res.status(401).end(ANONYMOUS_USER_NOT_AUTHORIZED);
```
zognin: should this be 403?

```suggestion
      return res.status(403).end(ANONYMOUS_USER_NOT_AUTHORIZED);
```

utils/auth/jwtHelpers.ts
```diff
@@ -0,0 +1,105 @@
...
+    // Single Authorization header.
+    bearerToken = authHeader.startsWith(bearerPrefix) ? authHeader : undefined;
+  }
+
+  if (!bearerToken) {
+    throw new Error('Bearer token not found in request headers.');
+  }
+
+  return bearerToken.replace(bearerPrefix, '');
+}
```
zognin: I think we can use guard clauses and change the wording a bit to make it clearer that we want an auth header that is a bearer token.

Also we can reduce the use of strings since there's an `.authorization` property. It's a `string | undefined` by compile time, so we don't need to check the type.

Also I would rather use control flow than throw exceptions because we are throwing general errors which may make it hard to trace down the line. But I think we don't need to focus on this

```suggestion
function getBearerToken(req: NextApiRequest) {
  const bearerPrefix = 'Bearer ';
  const authHeader = req.headers?.authorization;

  if (!authHeader) {
    throw new Error('Auth header is not present');
  }

  if (!authHeader.startsWith(bearerPrefix)) {
    throw new Error('Auth header is not a bearer token.');
  }

  return authHeader.replace(bearerPrefix, '');
}
```

utils/auth/jwtHelpers.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return handler(uid, req, res);
+  };
+}
+
+interface UserDetailsFromRequest {
+  uid: string;
+  isAnonymous: boolean;
+}
```
zognin: Should we move types and interfaces to the top of the file or a separate file? For easy finding. Right now it's used both in the function above and below it.

utils/auth/jwtHelpers.ts
zognin, line 66:
```suggestion
      return res.status(403).end(ANONYMOUS_USER_NOT_AUTHORIZED);
```

zognin (commented):
The functionalities work well

bharathcs:
Thanks @zognin, the suggestions all make sense and I've made the changes.

utils/auth/jwtHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+  });
+} catch (error) {
+  // Ignore the "already exists" message in dev environments (cause by hot reloads).
+  if (!(process.env.NODE_ENV == 'development' && /already exists/u.test('' + error))) {
+    console.error(error);
+  }
+}
+
+const UNABLE_TO_AUTHENTICATE = 'User could not be authenticated.';
+const ANONYMOUS_USER_NOT_AUTHORIZED = 'User has not verified their identity.';
```
zognin: Nit, wonder if the option below is more concise, it also does not limit "not authorized" to anonymous users
```suggestion
const USER_NOT_AUTHORIZED = 'User has not verified their identity.';
```

zognin (approved):
LGTM, you may merge it if you find it okay because I put one nit

PR #12, opened 2022-09-07
Add some API routes

sivayogasubramanian:
Added:

- /applications/:applicationId
  - GET
  - DELETE  
- /applications/:applicationId/tasks
  - POST
- /roles
  - LIST
  - POST

From this PR all API responses will have the same shape:

```ts
{
  "payload": ...data...,
  "messages": ...status messages...
}
```


zognin:
@sivayogasubramanian I know this is still a draft but I have some questions

- Is it necessary to state allowed http methods in the header? Because I think it's very easy to forget to check the default block when you are just trying to add new cases. Is it not enough to know that the request you are sending right now is not an allowed method?
- Why is every http util function in a different file?

types/apiResponse.ts
```diff
@@ -0,0 +1,18 @@
+export enum StatusMessageType {
+  Error = 0,
+  Warning = 1,
+  Information = 2,
```
zognin: Should we remove these if they're not used?
```suggestion
```

types/apiResponse.ts
```diff
@@ -0,0 +1,18 @@
+export enum StatusMessageType {
+  Error = 0,
```
zognin: Should these be string enums instead of numbers? The status message type shows up as a number which isn't very clear.
Screenshot 2022-09-08 at 9 31 34 PM

pages/api/applications/[applicationId].ts
```diff
@@ -1,5 +1,112 @@
...
+    res
+      .status(HTTP_STATUS_UNAUTHORIZED)
+      .json(createJsonResponse({}, messages.get(MessageType.APPLICATION_DOES_NOT_BELONG_TO_USER)));
+    return;
+  }
+
+  res.status(HTTP_STATUS_OK).json(createJsonResponse({}, messages.get(MessageType.APPLICATION_DELETED_SUCCESSFULLY)));
 }
+
+export default withVerifiedUser(handler);
```
Comment (zognin) [inline pages/api/applications/[applicationId].ts]: Anonymous users can do everything except access `roles/world` stuff. I think you need to change other files too.
```suggestion
export default withAnyUser(handler);
```

pages/api/applications/[applicationId].ts
```diff
@@ -1,5 +1,112 @@
...
+  userId: string,
+  req: NextApiRequest,
+  res: NextApiResponse<ApiResponse<ApplicationData | EmptyPayload>>,
+) {
+  const applicationId = Number(req.query.applicationId);
+
+  const application: ApplicationData | null = await prisma.application.findFirst({
+    where: {
+      id: applicationId,
+      userId,
```
Comment (zognin) [inline pages/api/applications/[applicationId].ts:59]: Should we be explicit? Since everything else is explicit.
```suggestion
      userId: userId,
```

Also the error message is a bit funny because it says NOT_FOUND when the application doesn't belong to the user. But I can't think of a better way around this rn.

types/apiResponse.ts
```diff
@@ -0,0 +1,18 @@
...
+  Warning = 1,
+  Information = 2,
+  Success = 3,
```
bharathcs: could this be a string enum instead? (lets us return a useful string when logging / returning via API instead of 0...3)
```suggestion
  Error = 'ERROR',
  Warning = 'WARN',
  Information = 'INFO',
  Success = 'SUCCESS',
```

pages/api/applications/[applicationId].ts
```diff
@@ -1,5 +1,112 @@
...
+}
+
+async function handleGet(
+  userId: string,
+  req: NextApiRequest,
+  res: NextApiResponse<ApiResponse<ApplicationData | EmptyPayload>>,
+) {
+  const applicationId = Number(req.query.applicationId);
+
+  const application: ApplicationData | null = await prisma.application.findFirst({
```
Comment (zognin) [inline pages/api/applications/[applicationId].ts]: If we handle prisma errors for `.create`, why don't we catch for other operations like `findFirst`?

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -1,5 +1,128 @@
...
+  ],
+  [
+    MessageType.TASK_APPLICATION_NOT_FOUND,
+    [{ type: StatusMessageType.Error, message: 'Application for this task cannot be found.' }],
+  ],
+  [
+    MessageType.TASK_APPLICATION_DOES_NOT_BELONG_TO_USER,
+    [{ type: StatusMessageType.Error, message: 'Application for this task does not belong to you.' }],
+  ],
+  [MessageType.TASK_CREATE_SUCCESSFULLY, [{ type: StatusMessageType.Success, message: 'Task created succesfully.' }]],
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/index.ts]: ?
```suggestion
  [MessageType.TASK_CREATED_SUCCESSFULLY, [{ type: StatusMessageType.Success, message: 'Task created succesfully.' }]],
```

utils/http/httpHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Utility function that rejects all responses not in allowed methods with a 405 HTTP response.
+ *
+ * @param res The NextApiResponse object.
+ * @param method The HTTP method that is rejected.
+ */
+export function rejectHttpMethod(res: NextApiResponse, method?: string) {
+  res
+    .status(405)
+    .json(createJsonResponse({}, [{ type: StatusMessageType.Error, message: `HTTP method ${method} not allowed!` }]));
```
bharathcs: Super nit: I'd rather hide the undefined in `HTTP method undefined not allowed!`.

```suggestion
    .json(createJsonResponse({}, [{ type: StatusMessageType.Error, message: `HTTP method ${method ? method + " " : ""}not allowed!` }]));
```

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -1,5 +1,128 @@
...
+      .json(createJsonResponse({}, messages.get(MessageType.TASK_APPLICATION_NOT_FOUND)));
+    return;
+  }
+
+  if (application.userId !== userId) {
+    res
+      .status(HTTP_STATUS_FORBIDDEN)
+      .json(createJsonResponse({}, messages.get(MessageType.TASK_APPLICATION_DOES_NOT_BELONG_TO_USER)));
+    return;
+  }
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/index.ts]: I feel like this check should be above the validation of request fields. If the user is unauthorized or the application doesn't exist, we shouldn't even bother validating the request.

utils/http/httpHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export function createJsonResponse<D>(payload: D | EmptyPayload, messages?: StatusMessage[]): ApiResponse<D> {
+  return {
+    payload,
+    messages: messages ?? [],
+  };
+}
```
bharathcs: Another really minor nit: I feel like the most common use case would actually be a single message, rather than an array of messages. Can we support passing in a single message? We can check the type and wrap it in an array within this function.

```suggestion
export function createJsonResponse<D>(
  payload: D | EmptyPayload,
  messages?: StatusMessage | StatusMessage[],
): ApiResponse<D> {
  // If messages is a single message, wrap it as an array of 1.
  if (messages && 'message' in messages) {
    messages = [messages];
  }

  return {
    payload,
    messages: messages ?? [],
  };
}
```

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -1,5 +1,128 @@
...
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.EMPTY_TITLE)));
+    return false;
+  }
+
+  if (!isValidDate(req.body.dueDate)) {
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.INVALID_DUE_DATE)));
+    return false;
+  }
+
+  if (req.body.notificationDateTime && !isValidDate(req.body.notificationDateTime)) {
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/index.ts]: I can pass in empty string and the request will work. It makes the field null.
```suggestion
  if (req.body.notificationDateTime !== undefined && !isValidDate(req.body.notificationDateTime)) {
```

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+  ROLE_CREATED_SUCCESSFULLY,
+}
+
+const messages = new Map<MessageType, StatusMessage[]>([
+  [
+    MessageType.INVALID_COMPANY_ID,
+    [{ type: StatusMessageType.Error, message: 'The company for this role is invalid.' }],
+  ],
+  [
+    MessageType.COMPANY_DOES_NOT_EXISTS,
```
zognin:
```suggestion
    MessageType.COMPANY_DOES_NOT_EXIST,
```

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+// TODO: Optional Search query. Will be done in another PR to unblock others.
+
+async function handleGet(_: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
+  const roles: RoleListData[] = await prisma.role.findMany({
+    select: {
+      id: true,
+      title: true,
+      type: true,
+      year: true,
+      isVerified: true,
```
zognin, line 70: Are we supposed to filter for only verified roles? Right now it returns unverified roles too.

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+
+    res
+      .status(HTTP_STATUS_CREATED)
+      .json(createJsonResponse(newRole, messages.get(MessageType.ROLE_CREATED_SUCCESSFULLY)));
+  });
+}
+
+async function isValidRequest(req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>): Promise<boolean> {
+  if (typeof req.body.companyId !== 'number') {
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.INVALID_COMPANY_ID)));
```
zognin: Maybe this should say "Company id must be a number"

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+    return false;
+  }
+
+  const company = await prisma.company.findFirst({ where: { id: req.body.companyId } });
+
+  if (!company) {
```
zognin: Why do you handle the similar case in `/tasks/index.ts` as `HTTP_STATUS_NOT_FOUND`, but you handle this as `HTTP_STATUS_BAD_REQUEST`. Also why is the former in the handler but the later in the validation function? Can we choose 1 way?

pages/api/applications/[applicationId].ts
```diff
@@ -1,5 +1,112 @@
...
+    return;
+  }
+
+  res.status(HTTP_STATUS_OK).json(createJsonResponse(application));
+}
+
+async function handleDelete(userId: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
+  const applicationId = Number(req.query.applicationId);
+
+  const { count } = await prisma.application.deleteMany({ where: { id: applicationId, userId } });
```
Comment (bharathcs) [inline pages/api/applications/[applicationId].ts]: Minor nit to explain the control flow lower down.

```suggestion

  // As applicationId is a primary key, count will only match 0 or 1. This is effectively 'delete if exists' behavior.
  const { count } = await prisma.application.deleteMany({ where: { id: applicationId, userId } });
```

pages/api/applications/[applicationId].ts
```diff
@@ -1,5 +1,112 @@
...
+      rejectHttpMethod(res, method);
+  }
+}
+
+async function handleGet(
+  userId: string,
+  req: NextApiRequest,
+  res: NextApiResponse<ApiResponse<ApplicationData | EmptyPayload>>,
+) {
+  const applicationId = Number(req.query.applicationId);
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId].ts:54]: Should we check if `req.query.applicationId` can be parsed to a number?

pages/api/applications/[applicationId].ts
Comment (ganhongyao) [inline pages/api/applications/[applicationId].ts]:
```suggestion
  const application: Nullable<ApplicationData> = await prisma.application.findFirst({
```
Perhaps we could define a `Nullable<T>` type that can be reused?

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -1,5 +1,128 @@
...
+      select: { id: true, title: true, dueDate: true, notificationDateTime: true, isDone: true },
+    });
+
+    res
+      .status(HTTP_STATUS_CREATED)
+      .json(createJsonResponse(newTask, messages.get(MessageType.TASK_CREATE_SUCCESSFULLY)));
+  });
+}
+
+function isValidRequest(req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>): boolean {
```
Comment (bharathcs) [inline pages/api/applications/[applicationId]/tasks/index.ts]: `isValidRequest` implies it only checks and returns bool. Its also better practice to handle the res in the handler function (for clarity). Could we return the messageType if it fails or undefined if it passes, then use it in the handler?


```suggestion
function validateRequestOrReject(req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>): MessageType | undefined {
```

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.EMPTY_TITLE)));
+    return false;
+  }
+
+  if (isEmpty(req.body.type) || !Object.values(RoleType).includes(req.body.type)) {
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.INVALID_TYPE)));
+    return false;
+  }
+
+  if (typeof req.body.year !== 'number') {
```
zognin, line 119: Right now any number is okay?

utils/auth/jwtHelpers.ts
```diff
@@ -56,25 +60,54 @@ export function withAnyUser<T>(
...
+/**
+ * Utility function that takes care of authenticating any user (anonymous / verified).
+ *
+ * Only calls the provided handler if the user has been successfully authenticated. Returns a 401 Unauthorized response
+ * if requirements are not met.
+ *
+ * @param handler Takes in an API handler function that needs an authenticated user.
+ * @returns NextApiHandler function.
+ */
+export function withAuthUser<T>(
```
zognin: I'm confused because different things are named similarly.
```suggestion
export function withAuth<T>(
```

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -1,5 +1,128 @@
...
+  }
+
+  if (application.userId !== userId) {
+    res
+      .status(HTTP_STATUS_FORBIDDEN)
+      .json(createJsonResponse({}, messages.get(MessageType.TASK_APPLICATION_DOES_NOT_BELONG_TO_USER)));
+    return;
+  }
+
+  createIfPossible(res, async () => {
```
Comment (bharathcs) [inline pages/api/applications/[applicationId]/tasks/index.ts]: I don't think its good practice to send response body outside of the handler function, and it shouldn't leave the module altogether. Can we have something like this:

```
const createdOrError = createIfPossible(async() => { ... } );
if (createdOrError instanceof Error) {
  res.send(errorCode).json(errorsMsgs);
}
// ...
```

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
 }
+
+async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleData | EmptyPayload>>) {
+  if (!(await isValidRequest(req, res))) {
+    return;
+  }
+
+  const rolePostData: RolePostData = req.body;
+
+  createIfPossible(res, async () => {
```
bharathcs: Mentioned elsewhere, I recommend not passing in res to another module.

utils/prisma/prismaHelpers.ts
```diff
@@ -0,0 +1,47 @@
...
+import { StatusMessageType } from '../../types/apiResponse';
+import { createJsonResponse, HTTP_STATUS_BAD_REQUEST, HTTP_STATUS_INTERNAL_SERVER_ERROR } from '../http/httpHelper';
+
+/**
+ * Utility function that takes care of Prisma runtime errors during CREATE.
+ *
+ * @param res The NextApiResponse object.
+ * @param creator The Prisma Creator function.
+ */
+export function createIfPossible(res: NextApiResponse, creator: () => Promise<void>) {
```
bharathcs: Mentioned elsewhere, I recommend not taking in the response and modifying it outside the handler. Only middleware / handlers should directly work on it, IMHO.

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
+import { isEmpty } from '@firebase/util';
```
bharathcs: Sorry, did you mean this?

```suggestion
import { isEmpty } from '../../../../../utils/strings/validations';
```

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+      select: { id: true, title: true, type: true, year: true },
+    });
+
+    res
+      .status(HTTP_STATUS_CREATED)
+      .json(createJsonResponse(newRole, messages.get(MessageType.ROLE_CREATED_SUCCESSFULLY)));
+  });
+}
+
+async function isValidRequest(req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>): Promise<boolean> {
```
bharathcs: Mentioned elsewhere, prefer handling res inside original + isValidRequest sounds like it only checks and returns boolean, not make destructive actions.

types/apiResponse.ts
```diff
@@ -0,0 +1,18 @@
...
+  type: StatusMessageType;
+  message: string;
+}
+
+export interface ApiResponse<T> {
+  payload: T | EmptyPayload;
+  messages: StatusMessage[];
+}
+
+export type EmptyPayload = Record<string, never>;
```
zognin, line 16: do you need to specify `ApiResponse<T | EmptyPayload>` in every file, considering you already made `payload: T | EmptyPayload`?

bharathcs, line 16: Ooh what's this and how is it different from `{}`? (Just curious)

pages/api/applications/[applicationId].ts
```diff
@@ -1,5 +1,112 @@
...
+  [
+    MessageType.APPLICATION_DOES_NOT_BELONG_TO_USER,
+    [{ type: StatusMessageType.Error, message: 'Application does not belong to you.' }],
+  ],
+  [
+    MessageType.APPLICATION_DELETED_SUCCESSFULLY,
+    [{ type: StatusMessageType.Success, message: 'Application was deleted successfully.' }],
+  ],
+]);
+
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId].ts]:
```suggestion
const messages = Object.freeze({
  [MessageType.APPLICATION_NOT_FOUND]: { type: StatusMessageType.Error, message: 'Application cannot be found.' },
  [MessageType.APPLICATION_DOES_NOT_BELONG_TO_USER]: {
    type: StatusMessageType.Error,
    message: 'Application does not belong to you.',
  },
  [MessageType.APPLICATION_DELETED_SUCCESSFULLY]: {
    type: StatusMessageType.Success,
    message: 'Application was deleted successfully.',
  },
});

```

Do we need to use a map for this? Making it a map also makes it mutable.

utils/auth/jwtHelpers.ts
```diff
@@ -32,14 +34,16 @@ const USER_NOT_AUTHORIZED = 'User has not verified their identity.';
  * @returns NextApiHandler function.
  */
 export function withAnyUser<T>(
```
zognin: since we have `withAuth`, we can rename this too, to make the meaning of `Auth` clear
```suggestion
export function withAuthUser<T>(
```

utils/http/httpHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+export const HTTP_GET_METHOD = 'GET';
+export const HTTP_POST_METHOD = 'POST';
+export const HTTP_PUT_METHOD = 'PUT';
+export const HTTP_DELETE_METHOD = 'DELETE';
+export const HTTP_PATCH_METHOD = 'PATCH';
+export const HTTP_HEAD_METHOD = 'HEAD';
+
+export const HTTP_STATUS_OK = 200;
+export const HTTP_STATUS_CREATED = 201;
+export const HTTP_STATUS_ACCEPTED = 202;
```
zognin: Can we remove the unused constants?

utils/http/httpHelper.ts
```diff
@@ -0,0 +1,40 @@
+import { NextApiResponse } from 'next';
+import { ApiResponse, EmptyPayload, StatusMessage, StatusMessageType } from '../../types/apiResponse';
+
```
zognin: I think this file can be called `httpHelpers.ts`? It's singular right now.

utils/http/httpHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param res The NextApiResponse object.
+ * @param method The HTTP method that is rejected.
+ */
+export function rejectHttpMethod(res: NextApiResponse, method?: string) {
+  res
+    .status(405)
+    .json(createJsonResponse({}, [{ type: StatusMessageType.Error, message: `HTTP method ${method} not allowed!` }]));
+}
+
+export function createJsonResponse<D>(payload: D | EmptyPayload, messages?: StatusMessage[]): ApiResponse<D> {
```
zognin: Why do we use `T` everywhere and suddenly this is `D`?

utils/auth/jwtHelpers.ts
```diff
@@ -56,25 +60,54 @@ export function withAnyUser<T>(
...
 
+/**
+ * Utility function that takes care of authenticating any user (anonymous / verified).
+ *
+ * Only calls the provided handler if the user has been successfully authenticated. Returns a 401 Unauthorized response
+ * if requirements are not met.
+ *
+ * @param handler Takes in an API handler function that needs an authenticated user.
+ * @returns NextApiHandler function.
+ */
```
bharathcs: Could you modify the docs to indicate that uid isn't returned + use withAnyUser if did is required.

utils/http/httpHelper.ts
```diff
@@ -0,0 +1,40 @@
...
+  return {
+    payload,
```
zognin: Can we do `payload: payload`? But actually up to you

utils/prisma/prismaHelpers.ts
```diff
@@ -0,0 +1,47 @@
...
+      );
+      return;
+    }
+
+    if (error instanceof Prisma.PrismaClientValidationError) {
+      res
+        .status(HTTP_STATUS_BAD_REQUEST)
+        .json(
+          createJsonResponse({}, [
+            { type: StatusMessageType.Error, message: 'Missing fields in request. Please try again.' },
```
zognin: This may not only be due to missing fields right? https://www.prisma.io/docs/reference/api-reference/error-reference#prismaclientvalidationerror

utils/auth/jwtHelpers.ts
```diff
@@ -56,25 +60,54 @@ export function withAnyUser<T>(
...
+      await getUserFromJwt(getBearerToken(req));
+    } catch (e) {
+      return res
+        .status(HTTP_STATUS_UNAUTHORIZED)
+        .json(createJsonResponse({}, [{ type: StatusMessageType.Error, message: UNABLE_TO_AUTHENTICATE }]));
+    }
+
+    return handler(req, res);
+  };
+}
```
bharathcs:
```suggestion
): NextApiHandler<ApiResponse<T | EmptyPayload>> {
  return withAnyUser((_, req, res) => handler(req, res));
}
```

utils/http/httpHelper.ts
ganhongyao:
```suggestion
export function createJsonResponse<D>(payload: D | EmptyPayload, ...messages: StatusMessage[]): ApiResponse<D> {
  return { payload, messages };
}
```

Would this be better since we don't have to pass in an array each time? Also, when will there be more than one message?

utils/strings/validations.ts
```diff
@@ -0,0 +1,3 @@
+export function isEmpty(value: string) {
+  return !value || !value.trim();
```
zognin, line 2: Just now we realised sending strings with only whitespaces through postman passes this check. It fails for console. I guess we need to figure out why.

zognin (commented):
I briefly tested the apis. I think most of them work okay, but there may be edge cases I haven't checked.

bharathcs (approved):
Tbh feels like a lot of my comments are nits LOL. It's not a problem if you disagree and want to resolve and move on.

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -1,5 +1,128 @@
...
+  }
+
+  const taskPostData: TaskPostData = {
+    ...req.body,
+    dueDate: new Date(req.body.dueDate),
+    notificationDateTime: req.body.notificationDateTime ? new Date(req.body.notificationDateTime) : null,
+  };
+
+  const applicationId = Number(req.query.applicationId);
+  const application = await prisma.application.findFirst({ where: { id: applicationId } });
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/index.ts]:
```suggestion
  const application = await prisma.application.findUnique({ where: { id: applicationId } });
```
Since id is a primary key, we should use `findUnique` instead of `findFirst`

pages/api/applications/[applicationId].ts
```diff
@@ -1,5 +1,112 @@
...
+  if (!application) {
+    res.status(HTTP_STATUS_NOT_FOUND).json(createJsonResponse({}, messages.get(MessageType.APPLICATION_NOT_FOUND)));
+    return;
+  }
+
+  res.status(HTTP_STATUS_OK).json(createJsonResponse(application));
+}
+
+async function handleDelete(userId: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
+  const applicationId = Number(req.query.applicationId);
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId].ts:107]: Same comment as above on number parsing

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -1,5 +1,128 @@
...
+    return;
+  }
+
+  const taskPostData: TaskPostData = {
+    ...req.body,
+    dueDate: new Date(req.body.dueDate),
+    notificationDateTime: req.body.notificationDateTime ? new Date(req.body.notificationDateTime) : null,
+  };
+
+  const applicationId = Number(req.query.applicationId);
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/index.ts:68]: Same comment as above on number parsing

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+  if (typeof req.body.companyId !== 'number') {
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.INVALID_COMPANY_ID)));
+    return false;
+  }
+
+  const company = await prisma.company.findFirst({ where: { id: req.body.companyId } });
```
ganhongyao: Same comment as above on using `findUnique`

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+} from '../../../utils/http/httpHelper';
+import { createIfPossible } from '../../../utils/prisma/prismaHelpers';
 
-export default function handler(req: NextApiRequest, res: NextApiResponse) {
-  res.status(200).end(`Request sent to ${req.url}\n`);
+const prisma = new PrismaClient();
+
+enum MessageType {
+  INVALID_COMPANY_ID,
+  COMPANY_DOES_NOT_EXISTS,
```
```suggestion
  COMPANY_DOES_NOT_EXIST,
```
Minor grammar nit

pages/api/applications/[applicationId].ts
Comment (bharathcs) [reply pages/api/applications/[applicationId].ts]: create will throw an error if we send bad data. findFirst will not throw an error for bad data but just null, which is handled later on in this function.

I believe we are catching for bad input data from the req. Other errors created by prism apps are stuff like database connection failures etc, which should probably have us return a 500, you're not wrong. There's a lot of places to modify this, so idm if this PR goes in without catching db conn failures and other 'unexpected' prisma failures.

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+    MessageType.INVALID_COMPANY_ID,
+    [{ type: StatusMessageType.Error, message: 'The company for this role is invalid.' }],
+  ],
+  [
+    MessageType.COMPANY_DOES_NOT_EXISTS,
+    [{ type: StatusMessageType.Error, message: 'The company for this role does not exists.' }],
+  ],
+  [MessageType.EMPTY_TITLE, [{ type: StatusMessageType.Error, message: 'Role title is empty.' }]],
+  [MessageType.INVALID_TYPE, [{ type: StatusMessageType.Error, message: 'Role type is invalid.' }]],
+  [MessageType.INVALID_YEAR, [{ type: StatusMessageType.Error, message: 'Role year is invalid.' }]],
```
ganhongyao: Could we provide more descriptive status messages? For a role year to be invalid could mean many things other than not being a `number`

pages/api/applications/[applicationId]/tasks/index.ts
Comment (bharathcs) [reply pages/api/applications/[applicationId]/tasks/index.ts]: Would recommend checking for undefined OR null, not just undefined. E.g. if the client sends `{ notificationDateTime: null }` to mark as empty. (IIRC null is a valid http datatype)

pages/api/roles/index.ts
```diff
@@ -1,5 +1,129 @@
...
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.COMPANY_DOES_NOT_EXISTS)));
+    return false;
+  }
+
+  if (isEmpty(req.body.title)) {
+    res.status(HTTP_STATUS_BAD_REQUEST).json(createJsonResponse({}, messages.get(MessageType.EMPTY_TITLE)));
+    return false;
+  }
+
+  if (isEmpty(req.body.type) || !Object.values(RoleType).includes(req.body.type)) {
```
```suggestion
  if (isEmpty(req.body.type) || !(req.body.type in RoleType) {
```

prisma/seed.ts
```diff
@@ -2,7 +2,15 @@ import { Prisma, PrismaClient } from '@prisma/client';
 
 const prisma = new PrismaClient({ log: ['query'] });
 
-const users: Prisma.UserCreateInput[] = [{ uid: '1' }, { uid: '2' }, { uid: '3' }, { uid: '4' }, { uid: '5' }];
+// TODO: REVERT CHANGES HERE
```
ganhongyao: Is this a note to self?

types/apiResponse.ts
ganhongyao: Should we uppercase enums?

pages/api/roles/index.ts
sivayogasubramanian: Yes. I think @ganhongyao wanted didn't want to add a CHECK constraint so that people can track past applications.

utils/http/httpHelper.ts
sivayogasubramanian: D for Data. Not changing this.

utils/http/httpHelper.ts
```diff
@@ -0,0 +1,40 @@
+import { NextApiResponse } from 'next';
+import { ApiResponse, EmptyPayload, StatusMessage, StatusMessageType } from '../../types/apiResponse';
+
+export const HTTP_GET_METHOD = 'GET';
+export const HTTP_POST_METHOD = 'POST';
+export const HTTP_PUT_METHOD = 'PUT';
+export const HTTP_DELETE_METHOD = 'DELETE';
+export const HTTP_PATCH_METHOD = 'PATCH';
+export const HTTP_HEAD_METHOD = 'HEAD';
```
ganhongyao: Nit, but would it be better to make these enums like this? 
```typescript
export enum HttpMethod {
   GET = 'GET',
   POST = 'POST',
   ...
}
```

So that we can use it as `HttpMethod.GET` instead of `HTTP_GET_METHOD`.

Same for statuses below.

utils/http/httpHelper.ts
sivayogasubramanian: Why? If we need it later we have to add it back. I think its okay to have all in this file now. Not removing.

pages/api/applications/[applicationId].ts
Comment (bharathcs) [reply pages/api/applications/[applicationId].ts:59]: I think its idiomatic not to specify explicitly like that when the key matches variable name.

utils/prisma/prismaHelpers.ts
bharathcs: Could be due to bad type in field too, nice catch.

utils/prisma/prismaHelpers.ts
```diff
@@ -0,0 +1,47 @@
...
+export function createIfPossible(res: NextApiResponse, creator: () => Promise<void>) {
+  creator().catch((error) => {
+    console.error(error);
+
+    if (error instanceof Prisma.PrismaClientKnownRequestError) {
+      res.status(HTTP_STATUS_BAD_REQUEST).json(
+        createJsonResponse({}, [
+          {
+            type: StatusMessageType.Error,
+            message: 'A database error has occured due to the request. Please try again.',
```
ganhongyao:
```suggestion
            message: 'A database error has occured due to the request.',
```
Would asking them to try again help? Since it's a database error (which probably means there is some validation error or database constraint violation)  

Not sure if we can add more information to this error message (in future perhaps) since it implies that there is something the user can probably do to fix?

utils/strings/validations.ts
ganhongyao:
```suggestion
  return !value.trim();
```
Minor nit since this covers both cases

pages/api/applications/[applicationId].ts
Comment (sivayogasubramanian) [reply pages/api/applications/[applicationId].ts:59]: @bharathcs I think let's just change it. Sometimes my brain works slower because of this.

utils/prisma/prismaHelpers.ts
ganhongyao:
```suggestion
export function handlePrismaAction(res: NextApiResponse, prismaAction: () => Promise<void>) {
```
Is this function specific to `CREATE` actions? If it is then just ignore this comment.

pages/api/applications/[applicationId].ts
Comment (ganhongyao) [reply pages/api/applications/[applicationId].ts:59]: I agree with @bharathcs on this, it is quite idiomatic

pages/api/applications/[applicationId].ts
Comment (sivayogasubramanian) [reply pages/api/applications/[applicationId].ts:54]: We wouldn't hit this route if its not a number right?

utils/http/httpHelper.ts
sivayogasubramanian: Will be using @ganhongyao at https://github.com/sivayogasubramanian/caza/pull/12#discussion_r966069188

utils/prisma/prismaHelpers.ts
bharathcs: Not necessarily, could just be a db conn limit, in which case trying again (after a while) will work.

utils/http/httpHelper.ts
bharathcs: Nice! Much better than my solution...

pages/api/applications/[applicationId].ts
Comment (bharathcs) [reply pages/api/applications/[applicationId].ts]: Hmm why tho? I don't believe it adds more clarity / adds any behaviour unless u mean to implement Nullable like an Optional monad.

pages/api/applications/[applicationId].ts
Comment (bharathcs) [reply pages/api/applications/[applicationId].ts:54]: It can, e.g. `api/applications/foo` endpoint. We'll then get a prism client validation error:

error - unhandledRejection: PrismaClientValidationError: 
Invalid `prisma.application.findFirst()` invocation:

{
  where: {
    id: NaN,
        ~~~

types/apiResponse.ts
sivayogasubramanian: I used this at first then the linters forced me to change to this lol

pages/api/applications/[applicationId].ts
Comment (sivayogasubramanian) [reply pages/api/applications/[applicationId].ts:54]: @bharathcs I still don't get it. The application id is from the url query params. If it wasn't a number why would this function run?

prisma/seed.ts
sivayogasubramanian: Yes

pages/api/applications/[applicationId].ts
Comment (bharathcs) [reply pages/api/applications/[applicationId].ts:54]: Query parameters can be strings like "foo" right? Or is this some REST thing I'm not aware of?

Comment (sivayogasubramanian) [reply pages/api/applications/[applicationId].ts:54]: @bharathcs you are right. will fix it.

pages/api/roles/index.ts
zognin: But I can put -1 and that would also work?

utils/strings/validations.ts
bharathcs:
```suggestion
  return !value.trim();
```

bharathcs (approved):
LGTM!

pages/api/roles/index.ts
```diff
@@ -1,5 +1,130 @@
...
+
+  res.status(HttpStatus.CREATED).json(createJsonResponse(newRole, messages[MessageType.ROLE_CREATED_SUCCESSFULLY]));
+}
+
+function validateRequest(req: NextApiRequest): Nullable<MessageType> {
+  if (typeof req.body.companyId !== 'number') {
+    return MessageType.INVALID_COMPANY_ID;
+  }
+
+  if (req.body.title && isEmpty(req.body.title)) {
```
zognin, line 111: During create, `title` is a compulsory field.
This is different for patch, or for creating entities with non-compulsory fields.
You may want to check for the other fields as well.

```suggestion
 if (!req.body.title) {
   // Error that title is missing
 }

  if (isEmpty(req.body.title)) {
```
Btw, not including `req.body.title` leads to runtime errors. I'm not sure if `withPrismaErrorHandling` was supposed to catch that.

bharathcs: Resolved to be solved later: #21

PR #13, opened 2022-09-07
Add custom type guards to validate enum type and satisfy compiler type checks.

bharathcs:
Type guards are necessary to validate an object as a particular type and also to appease the compiler's type-checking.
For typescript compiler to accept an object as a enum value, typical `instanceof` or `typeof` solutions do not work.

See here for more details on how this type guard works: https://stackoverflow.com/questions/56044872/typescript-enum-object-values-return-value/56044932#56044932

Checkout the second commit to see how I tested it (did both runtime & compiler tests):

- Create custom type guards for the two enums.
- Test the custom type guard out.
- Revert "Test the custom type guard out."


bharathcs:
No longer necessary, we will rely on Prisma to do the model validation. (Discussed offline)

PR #14, opened 2022-09-08
Add companies api

zognin:
Handlers
- POST
  - Company (name, companyUrl).
  - companyUrl will be trimmed to remove protocol and www if present.
- GET
  - List of companies that are verified
  - Will only return companies whose names contain the search keyword, case insensitive. If the search keyword is not provided, it returns nothing.

Utils
- String formatters and validations

Database
- Added unique index for Company(name, companyUrl), **please drop db if necessary and regenerate with `yarn prisma migrate dev`**

### DIFF
--- a/pages/api/companies.ts
+++ b/pages/api/companies.ts
@@ -2,13 +2,14 @@ import { PrismaClient } from '@prisma/client';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { CompanyData, CompanyListData, CompanyPostData } from '../../types/company';
 import { isEmpty, isValidUrl } from '../../utils/strings/validations';
-import { removeProtocolAndWwwIfPresent } from '../../utils/strings/formatters';
+import { capitalizeEveryWord, removeProtocolAndWwwIfPresent } from '../../utils/strings/formatters';
 import { createJsonResponse, HttpMethod, HttpStatus, rejectHttpMethod } from '../../utils/http/httpHelpers';
 import { withAuth } from '../../utils/auth/jwtHelpers';
 import { ApiResponse, StatusMessageType } from '../../types/apiResponse';
 import { Nullable } from '../../types/utils';
 
 enum MessageType {
+  COMPANY_ALREADY_EXISTS,
   COMPANY_CREATED_SUCCESSFULLY,
   EMPTY_NAME,
   INVALID_COMPANY_URL,
@@ -19,6 +20,7 @@ enum MessageType {
 const prisma = new PrismaClient();
 
 const messages = Object.freeze({
+  [MessageType.COMPANY_ALREADY_EXISTS]: { type: StatusMessageType.SUCCESS, message: 'Company already exists.' },
   [MessageType.COMPANY_CREATED_SUCCESSFULLY]: {
     type: StatusMessageType.SUCCESS,
     message: 'Company created successfully.',
@@ -59,12 +61,29 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<
     res.status(HttpStatus.BAD_REQUEST).json(createJsonResponse({}, messages[errorMessageType]));
     return;
   }
-  const companyPostData: CompanyPostData = req.body;
+
+  const companyPostData: CompanyPostData = {
+    name: capitalizeEveryWord(req.body.name),
+    companyUrl: removeProtocolAndWwwIfPresent(req.body.companyUrl),
+  };
+
+  const duplicateCompany = await prisma.company.findFirst({
+    where: {
+      name: companyPostData.name,
+      companyUrl: companyPostData.companyUrl,
+    },
+    select: { id: true, name: true, companyUrl: true },
+  });
+
+  if (duplicateCompany) {
+    res.status(HttpStatus.OK).json(createJsonResponse(duplicateCompany, messages[MessageType.COMPANY_ALREADY_EXISTS]));
+    return;
+  }
 
   const newCompany = await prisma.company.create({
     data: {
       name: companyPostData.name,
-      companyUrl: removeProtocolAndWwwIfPresent(companyPostData.companyUrl),
+      companyUrl: companyPostData.companyUrl,
     },
     select: { id: true, name: true, companyUrl: true },
   });
--- a/pages/api/companies.ts
+++ b/pages/api/companies.ts
@@ -2,7 +2,7 @@ import { PrismaClient } from '@prisma/client';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { CompanyData, CompanyListData, CompanyPostData } from '../../types/company';
 import { isEmpty, isValidUrl } from '../../utils/strings/validations';
-import { capitalizeEveryWord, removeProtocolAndWwwIfPresent } from '../../utils/strings/formatters';
+import {capitalizeEveryWord, removeProtocolAndWwwIfPresent, trim} from '../../utils/strings/formatters';
 import { createJsonResponse, HttpMethod, HttpStatus, rejectHttpMethod } from '../../utils/http/httpHelpers';
 import { withAuth } from '../../utils/auth/jwtHelpers';
 import { ApiResponse, StatusMessageType } from '../../types/apiResponse';
@@ -49,7 +49,7 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
 
 async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<CompanyListData[]>>) {
   const { name } = req.query;
-  const trimmedName = name && name.toString().trim();
+  const trimmedName = name && trim(name);
 
   if (!trimmedName) {
     res.status(HttpStatus.OK).json(createJsonResponse([]));
--- a/pages/api/companies.ts
+++ b/pages/api/companies.ts
@@ -49,9 +49,9 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
 
 async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<CompanyListData[]>>) {
   const { name } = req.query;
-  const trimmedNameKeyword = name && name.toString().trim();
+  const trimmedName = name && name.toString().trim();
 
-  if (!trimmedNameKeyword) {
+  if (!trimmedName) {
     res.status(HttpStatus.OK).json(createJsonResponse([]));
     return;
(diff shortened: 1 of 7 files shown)


pages/api/companies.ts
```diff
@@ -1,5 +1,130 @@
...
+    case HttpMethod.POST:
+      handlePost(req, res);
+      break;
+    default:
+      rejectHttpMethod(res, method);
+  }
+}
+
+async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<CompanyListData[]>>) {
+  if (req.body.searchKeyword === undefined || req.body.searchKeyword === null) {
```
bharathcs: If the searchKeyword isn't provided, shouldn't we return all companies? I also don't quite get why we are returning [] as the result.

Also, if the searchKeyword is empty or whitespace only, shouldn't we have the same behaviour as undefined searchKeyword?

pages/api/companies.ts
```diff
@@ -1,5 +1,130 @@
...
+  const searchKeyword = req.body.searchKeyword.trim();
+
+  const companies: CompanyListData[] = await prisma.company.findMany({
+    where: {
+      isVerified: true,
+      name: {
+        contains: searchKeyword,
+        mode: 'insensitive',
+      },
+    },
```
bharathcs: Do we want to support searching by company url as well? Smth like:

```suggestion
    where: {
      AND: [
        { isVerified: true },
       OR: [
         name: {
           contains: searchKeyword,
           mode: 'insensitive',
         },
         companyUrl: {
           contains: searchKeyword,
           mode: 'insensitive',
         }
       ] 
      ]
    },
```

types/company.ts
```diff
@@ -1,5 +1,12 @@
-export type CompanyData = {
+export type CompanyListData = {
   id: number;
   name: string;
   companyUrl: string;
 };
```
bharathcs, line 5: Super nit

```suggestion
export type CompanyListData = CompanyPostData & {
  id: number;
};
```

utils/strings/formatters.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (trimmedUrl.startsWith(httpsProtocolPrefix)) {
+    trimmedUrl = trimmedUrl.slice(httpsProtocolPrefix.length);
+  }
+
+  if (trimmedUrl.startsWith(wwwPrefix)) {
+    trimmedUrl = trimmedUrl.slice(wwwPrefix.length);
+  }
+
+  return trimmedUrl;
```
bharathcs, line 20: Nit: I think its simple enough for regex but if you wanna avoid it that's fine too.

```suggestion
  const pattern = /(http[s]?:\/\/)?(www.)?(.+)/;
  const match = url.match(pattern);
  return match.length === 4 ? match[3] : url;
```

utils/strings/formatters.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export function capitalizeEveryWord(phrase: string) {
+  return phrase
+    .toLowerCase()
+    .split(' ')
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(' ');
```
bharathcs, line 29: Woah I wasn't aware that .charAt(1) returns "" for single letter or zero letter strings. Could we add a comment to document it. (Or ignore it, this might just be me)
```suggestion
  return phrase
    .toLowerCase()
    .split(' ')
    // Safely handles word with length 0 or 1 as well.
    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
    .join(' ');
```

utils/strings/validations.ts
```diff
@@ -1,3 +1,14 @@
 export function isEmpty(value: string) {
   return !value.trim();
 }
+
+// Note: This denotes a url as valid even without a protocol or domain.
+export function isValidUrl(value: string) {
+  const patternWithProtocol = new RegExp(
+    '^(http[s]?:\\/\\/(www\\.)?|ftp:\\/\\/(www\\.)?|www\\.){1}([0-9A-Za-z-\\.@:%_+~#=]+)+((\\.[a-zA-Z]{2,3})+)(/(.)*)?(\\?(.)*)?',
```
bharathcs: Are we deliberately supporting subdirectories? e.g. `google.com/dslkfjasdlkfjsdkf/sdfkdjsf` will pass.

utils/strings/validations.ts
```diff
@@ -1,3 +1,14 @@
...
+  );
+
+  const patternWithoutProtocol = new RegExp('^([0-9A-Za-z-\\.@:%_+~#=]+)+((\\.[a-zA-Z]{2,3})+)(/(.)*)?(\\?(.)*)?');
```
bharathcs: If I'm reading this correctly, the TLD is specified as only being 2-3 characters right?

But I think `.bank`, `.google` and other TLDs are valid now

bharathcs (commented):
Looks good

pages/api/companies.ts
zognin: I think we will list companies only when people are trying to select them when creating an application? So they should already have a company in mind. So if they haven't typed anything in the bar, then we don't need to suggest any companies yet. I will update to return nothing for the empty string case.

But do you think we need to return all companies? Why?

pages/api/companies.ts
zognin: Same for the above comment, when will searching by url be useful? Users will only see the company name and want to find companies by name when making an application, they won't really see how the companyUrl is used in our app.

utils/strings/formatters.ts
zognin: I think I'm not adopting this in order to avoid regex as much as possible. Regex is less understandable than logic.

types/company.ts
sivayogasubramanian: I think we should not do this as the incoming POST data should not be tied to the outgoing response data.

pages/api/companies.ts
sivayogasubramanian: I agree with @zognin here.

utils/strings/validations.ts
zognin: When I tested it works even with longer TLD? Actually I'm also not so sure about regex :   ) Edit: I updated it to be 2 or more.

utils/strings/validations.ts
zognin: I fixed it to disallow subdirectories.

pages/api/companies.ts
```diff
@@ -1,5 +1,134 @@
...
+    select: { id: true, name: true, companyUrl: true },
+  });
+
+  res
+    .status(HttpStatus.CREATED)
+    .json(createJsonResponse(newCompany, messages[MessageType.COMPANY_CREATED_SUCCESSFULLY]));
+}
+
+function validateRequest(req: NextApiRequest): Nullable<MessageType> {
+  if (!req.body.name) {
```
sivayogasubramanian: Empty `req.body.name` will drop into this if statement. We need to do `req.body.name === undefined`

pages/api/companies.ts
```diff
@@ -1,5 +1,134 @@
...
+    return MessageType.MISSING_NAME;
+  }
+
+  if (isEmpty(req.body.name)) {
+    return MessageType.EMPTY_NAME;
+  }
+
+  if (!req.body.companyUrl) {
```
sivayogasubramanian: Same here. Will there be a isEmpty check for this?

pages/api/companies.ts
```diff
@@ -1,5 +1,134 @@
...
+      handlePost(req, res);
+      break;
+    default:
+      rejectHttpMethod(res, method);
+  }
+}
+
+async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<CompanyListData[]>>) {
+  const { name } = req.query;
+  const trimmedNameKeyword = name && name.toString().trim();
```
sivayogasubramanian: Let's extract this out into a string formatter too

pages/api/companies.ts
```diff
@@ -1,5 +1,134 @@
...
+  EMPTY_NAME,
+  INVALID_COMPANY_URL,
+  MISSING_NAME,
+  MISSING_COMPANY_URL,
 }
+
+const prisma = new PrismaClient();
+
+const messages = Object.freeze({
+  [MessageType.COMPANY_ALREADY_EXISTS]: { type: StatusMessageType.SUCCESS, message: 'Company already exists.' },
```
sivayogasubramanian, line 30:
```suggestion
  [MessageType.COMPANY_ALREADY_EXISTS]: { type: StatusMessageType.ERROR, message: 'Company already exists.' },
```

pages/api/companies.ts
```diff
@@ -1,5 +1,134 @@
...
+  COMPANY_CREATED_SUCCESSFULLY,
+  EMPTY_NAME,
+  INVALID_COMPANY_URL,
+  MISSING_NAME,
+  MISSING_COMPANY_URL,
 }
+
+const prisma = new PrismaClient();
+
+const messages = Object.freeze({
```
sivayogasubramanian, line 24: Lets follow the enum order for the order here too. ERROR first then SUCCESS then ...

pages/api/companies.ts
sivayogasubramanian:
```suggestion
  const trimmedName = name && name.toString().trim();
```

pages/api/companies.ts
sivayogasubramanian: Not sure if this is a success or error

utils/strings/validations.ts
```diff
@@ -1,3 +1,14 @@
 export function isEmpty(value: string) {
   return !value.trim();
 }
+
+// Note: This denotes a url as valid even without a protocol or domain.
+export function isValidUrl(value: string) {
```
sivayogasubramanian, line 6: image

Should we allow this kind of url?

When i type `indeed.com` it goes there automatically. Users will just copy paste this.

I foresee a lot of errors if we require users to end the url with a TLD.

pages/api/companies.ts
zognin: It's a success because if companies are not verified, users will attempt to create them multiple times and that should be okay.

pages/api/companies.ts
zognin: No need to check isEmpty, it will be handled in isValidUrl

sivayogasubramanian (approved):

PR #15, opened 2022-09-08
Add stage api

zognin:
Handlers
- PATCH
  - Any of {type, date, emojiUnicodeHex, remark}.
  - Request body fields must be validated otherwise there could be runtime errors when converting from string to date etc.
- DELETE

Some changes to previous standards

- Changed `validateRequest` to `validateMethodRequest` because different methods of the same endpoint can have different validations
- Decided to use `XX_METHOD_UNAUTHORIZED` as `MessageType` and similarly for the message, instead of saying it does not belong to user because `belonging to the user !== authorized`
- Moved validation of query prams into `validateMethodRequest` function

### DIFF
--- a/pages/api/applications/[applicationId]/stages/[stageId].ts
+++ b/pages/api/applications/[applicationId]/stages/[stageId].ts
@@ -185,14 +185,14 @@ function validatePatchRequest(req: NextApiRequest) {
     return MessageType.INVALID_APPLICATION_STAGE_TYPE;
   }
 
-  if (req.body.date !== undefined && !isValidDate(req.body.date)) {
+  if (req.body.date !== undefined && (typeof req.body.date !== 'string' || !isValidDate(req.body.date))) {
     return MessageType.INVALID_DATE;
   }
 
   if (
     req.body.emojiUnicodeHex !== undefined &&
     req.body.emojiUnicodeHex !== null &&
-    !isValidHex(req.body.emojiUnicodeHex)
+    (typeof req.body.emojiUnicodeHex !== 'string' || !isValidHex(req.body.emojiUnicodeHex))
   ) {
     return MessageType.INVALID_EMOJI_UNICODE_HEX;
   }
--- a/pages/api/applications/[applicationId]/stages/[stageId].ts
+++ b/pages/api/applications/[applicationId]/stages/[stageId].ts
@@ -39,10 +39,6 @@ const messages = Object.freeze({
     type: StatusMessageType.ERROR,
     message: 'Application stage cannot be found.',
   },
-  [MessageType.EMPTY_UPDATE]: {
-    type: StatusMessageType.ERROR,
-    message: 'At least one field must be provided for an update.',
-  },
   [MessageType.APPLICATION_STAGE_UPDATE_UNAUTHORIZED]: {
     type: StatusMessageType.ERROR,
     message: 'Application stage cannot be updated by the user.',
@@ -179,10 +175,6 @@ function validatePatchRequest(req: NextApiRequest) {
 
   if (!isInteger(req.query.applicationId as string)) {
     return MessageType.INVALID_APPLICATION_ID;
-  }
-
-  if (!req.body || Object.keys(req.body).length == 0) {
-    return MessageType.EMPTY_UPDATE;
   }
 
   if (req.body.type !== undefined && !(req.body.type in ApplicationStageType)) {
--- a/pages/api/applications/[applicationId]/stages/[stageId].ts
+++ b/pages/api/applications/[applicationId]/stages/[stageId].ts
@@ -16,6 +16,7 @@ enum MessageType {
   APPLICATION_STAGE_NOT_FOUND,
   APPLICATION_STAGE_UPDATE_UNAUTHORIZED,
   APPLICATION_STAGE_UPDATED_SUCCESSFULLY,
+  EMPTY_UPDATE,
   INVALID_APPLICATION_ID,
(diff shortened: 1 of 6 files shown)


types/applicationStage.ts
```diff
@@ -1,10 +1,19 @@
...
   remark: Nullable<string>;
 };
+
+export type ApplicationStageData = ApplicationStageApplicationData & { applicationId: number };
+
+export type ApplicationStagePatchData = {
+  type: ApplicationStageType;
+  date: Date;
+  emojiUnicodeHex: string | null;
+  remark: string | null;
```
sivayogasubramanian: Lets use Nullable here

pages/api/applications/[applicationId]/stages/[stageId].ts
```diff
@@ -1,5 +1,215 @@
...
+    return MessageType.INVALID_APPLICATION_STAGE_TYPE;
+  }
+
+  if (req.body.date !== undefined && !isValidDate(req.body.date)) {
+    return MessageType.INVALID_DATE;
+  }
+
+  if (
+    req.body.emojiUnicodeHex !== undefined &&
+    req.body.emojiUnicodeHex !== null &&
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/[stageId].ts:190]:
```suggestion
    req.body.emojiUnicodeHex != null 
```

pages/api/applications/[applicationId]/stages/[stageId].ts
```diff
@@ -1,5 +1,215 @@
...
+function validatePatchRequest(req: NextApiRequest) {
+  if (!isInteger(req.query.stageId as string)) {
+    return MessageType.INVALID_APPLICATION_STAGE_ID;
+  }
+
+  if (!isInteger(req.query.applicationId as string)) {
+    return MessageType.INVALID_APPLICATION_ID;
+  }
+
+  if (req.body.type !== undefined && !Object.values(ApplicationStageType).includes(req.body.type)) {
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/[stageId].ts]:
```suggestion
  if (req.body.type !== undefined && !(req.body.type in ApplicationStageType) {
```

pages/api/applications/[applicationId]/stages/[stageId].ts
```diff
@@ -1,5 +1,215 @@
...
+      break;
+    case HttpMethod.DELETE:
+      handleDelete(userId, req, res);
+      break;
+    default:
+      rejectHttpMethod(res, method);
+  }
+}
+
+async function handlePatch(
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/[stageId].ts:81]: Sending a `{}` json object to the PATCH request causes a runtime error.

I think the prisma error handling is not working 😔😔.

types/applicationStage.ts
```diff
@@ -1,10 +1,19 @@
...
   id: number;
   type: ApplicationStageType;
   date: Date;
   emojiUnicodeHex: Nullable<string>;
   remark: Nullable<string>;
 };
+
+export type ApplicationStageData = ApplicationStageApplicationData & { applicationId: number };
+
+export type ApplicationStagePatchData = {
```
sivayogasubramanian, line 14: I have not used PATCH before but should all the fields be optional?

pages/api/applications/[applicationId]/stages/[stageId].ts
```diff
@@ -1,5 +1,215 @@
...
+  if (errorMessageType != null) {
+    res.status(HttpStatus.BAD_REQUEST).json(createJsonResponse({}, messages[errorMessageType]));
+    return;
+  }
+
+  const applicationStageId = Number(req.query.stageId);
+  const applicationId = Number(req.query.applicationId);
+
+  const applicationStagePatchData: ApplicationStagePatchData = {
+    ...req.body,
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/[stageId].ts]: Doing this is dangerous. I can have some other field in req.body and it would be inside applicationStagePatchData now.

Although it does not matter for this case, i think its better for us to manually assign it.

pages/api/applications/[applicationId]/stages/[stageId].ts
```diff
@@ -1,5 +1,215 @@
...
+  }
+}
+
+async function handlePatch(
+  userId: string,
+  req: NextApiRequest,
+  res: NextApiResponse<ApiResponse<ApplicationStageData>>,
+) {
+  const errorMessageType = validatePatchRequest(req);
+  if (errorMessageType != null) {
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/[stageId].ts]:
```suggestion
  if (errorMessageType !== null) {
```

pages/api/applications/[applicationId]/stages/[stageId].ts
Comment (zognin) [reply pages/api/applications/[applicationId]/stages/[stageId].ts:190]: Not changing for clarity.

pages/api/applications/[applicationId]/stages/[stageId].ts
Comment (bharathcs) [reply pages/api/applications/[applicationId]/stages/[stageId].ts:81]: Alright I found the issue (related to #28). So these are the fixes, in order:

1. The `handler()` function returns before the async is completed: merge #30
2. The middleware from `jwtHelpers` and `prismaHelpers` return before the async is handled: merge #28
3.  The Prisma helper didn't account for this as a bad parameter rather than an internal server error: merge #31

For (1), I think we need to let everyone know that if you are using async handlers, make sure your entire function finishes before you return.

zognin:
Btw, assigning undefined will cause prisma update to fail. That's why I only assign the object properties if defined.

sivayogasubramanian (approved):
LGTM!

PR #16, opened 2022-09-08
Add applications api

zognin:
Handlers
- POST
  - Application (roleId, userId).
  - Does not check if roleId is a valid foreign key, left to prisma to handle.
- GET
  - List of applications and its relevant data. Refer to the types for more info.
  - Sorts the list by most taskNotifications, followed by latest date of latest stage.

**Note: Please run `yarn prisma generate` to regenerate the client, otherwise filtered relations count cannot compile**
References: 
- https://www.prisma.io/docs/concepts/components/prisma-client/aggregation-grouping-summarizing#filter-the-relation-count
- https://www.prisma.io/docs/concepts/components/preview-features/client-preview-features#enabling-a-prisma-client-preview-feature

Other notes
- Will do search in a different PR. May want to reuse some of the roles search logic.

### DIFF
--- a/pages/api/applications/index.ts
+++ b/pages/api/applications/index.ts
@@ -1,5 +1,5 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
-import { ApplicationPostData } from '../../../types/applications';
+import { ApplicationListData, ApplicationPostData } from '../../../types/application';
 import { PrismaClient } from '@prisma/client';
 import { ErrorData } from '../../../types/error';
 import { HttpMethod, HttpStatus, rejectHttpMethod } from '../../../utils/http/httpHelpers';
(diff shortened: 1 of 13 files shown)


utils/constants.ts
```diff
@@ -1 +1,2 @@
 export const MIN_ROLE_YEAR = 1970;
+export const MIN_DATE = new Date(-8640000000000000);
```
sivayogasubramanian, line 4: Let's add a comment here.

pages/api/applications/index.ts
```diff
@@ -1,5 +1,174 @@
...
+  }
+
+  if (req.body.roleId === null || !Number.isInteger(req.body.roleId)) {
+    return MessageType.INVALID_ROLE_ID;
+  }
+
+  return null;
+}
+
+export default withAuthUser(handler);
```
sivayogasubramanian: Add prisma error handling

types/application.ts
```diff
@@ -2,10 +2,29 @@ import { ApplicationStageApplicationData } from './applicationStage';
...
 };
+
+export type ApplicationRoleData = {
+  id: number;
+  role: RoleData & { company: CompanyData };
+};
+
+export type ApplicationPostData = {
+  roleId: number;
+  userId: string;
```
sivayogasubramanian: Why is there a userId here?

pages/api/applications/index.ts
```diff
@@ -1,5 +1,174 @@
...
+    .status(HttpStatus.CREATED)
+    .json(createJsonResponse(newApplication, messages[MessageType.APPLICATION_CREATED_SUCCESSFULLY]));
+}
+
+function validatePostRequest(req: NextApiRequest): Nullable<MessageType> {
+  if (req.body.roleId === undefined) {
+    return MessageType.MISSING_ROLE_ID;
+  }
+
+  if (req.body.roleId === null || !Number.isInteger(req.body.roleId)) {
```
sivayogasubramanian, line 167: Can you help to change all instances to `isInteger` to this?

sivayogasubramanian (approved):
LGTM! Pre-approving. Merge when the comments are resolved.

types/application.ts
zognin: Creation of application needs a userId, but I can also choose not to put it in here. Edit: I think I'll leave it there for clarity.

sivayogasubramanian: Isn't the userId available in the handler?

sivayogasubramanian: I feel like this should represent the shape of the JSON that we send over and not what prisma needs.

zognin: Okay

PR #20, opened 2022-09-09
Add POST, DELETE endpoints for '/api/user/'

bharathcs:
I tested with this plan (with the new seeds):

Run the step, check the status code, and check Prisma studio for the impact.

1. Try with bad method (neither POST nor DELETE) or without token.
2. Delete `devUserWithoutData`: 200, user deleted.
3. Delete `devUserDoesNotExist`: 400, no change
4. Post `devUserWithoutData`: 201, created
5. Post `devUserWithoutData`: 409, no change
6. Post `devUserWithoutData` (old_token: devUserNew): 401, no change
7. Post `devUserVerifiedWithData`(old_token: devUserWithoutData): 409, no change
8. Post `devUserVerifiedDoesNotExist` (old_token: devUserDoesNotExist): 404, no change.
9. Post `devUserVerifiedWithoutData` (old_token:  devUserWithData): 200, uid replaced and has old data
10. Delete `devUserVerifiedWithoutData`: 200, user deleted (note that it now contains data, but was deleted anw)


bharathcs:
Sorry forgot to push the uncommitted changes, everything is in now.

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+async function handlePostWithoutOldToken(
+  currentUid: string,
+  _req: NextApiRequest,
+  res: NextApiResponse<ApiResponse<EmptyPayload>>,
+) {
+  const isExisting = (await getUserIfExists(currentUid)) !== null;
+  if (isExisting) {
+    return res
+      .status(HttpStatus.CONFLICT)
+      .json(createJsonResponse({}, { type: ERROR, message: `UID ${currentUid} already exists` }));
```
zognin: I'm not sure if we want to expose the UID in the user facing message?

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  NEW_USER_LINKED,
+}
+
+const { ERROR, INFORMATION, SUCCESS } = StatusMessageType;
+const prisma = new PrismaClient();
+
+async function handler(currentUid: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
+  switch (req.method) {
+    case HttpMethod.POST:
+      const oldUserToken = req.body?.old_token?.trim();
```
zognin:
```suggestion
      const oldUserToken = req.body?.oldToken?.trim();
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      return oldUserToken /* if null, undefined, empty or contains only whitespace */
+        ? handlePostWithOldToken(currentUid, oldUserToken, req, res)
```
zognin: There's a runtime error and the app hangs when the token is not sent properly.

Screenshot 2022-09-10 at 12 50 40 PM

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  if (isExisting) {
+    return res
+      .status(HttpStatus.CONFLICT)
+      .json(createJsonResponse({}, { type: ERROR, message: `UID ${currentUid} already exists` }));
+  }
+
+  await prisma.user.create({ data: { uid: currentUid } });
+  return res
+    .status(HttpStatus.OK)
+    .json(createJsonResponse({}, { type: SUCCESS, message: `New user with UID ${currentUid} added.` }));
```
zognin: Not sure if you want to standardize putting messages in Object.freeze format, but I'm okay with this also.

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  return res
+    .status(HttpStatus.OK)
+    .json(createJsonResponse({}, { type: SUCCESS, message: `Deleted UID ${uid} and their data.` }));
+}
+
+async function linkAccount(oldUid: string, newUid: string): Promise<LinkAccountResult> {
+  const newUser = await getUserIfExists(newUid);
+  const oldUser = await getUserIfExists(oldUid);
+
+  // If old user has already been removed from the database. This can happen if the API was called twice or if the new user
```
zognin: I thought if `anonUid` is found and `verifiedUid` is not found, we are updating `anonUid` -> `verifiedUid`, without creating or deleting any old/new user?

I think this means the other cases are automatically handled.

This would allow cascading updates to happen so we can be assured the rest of our relations are correct?

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    return LinkAccountResult.OLD_USER_DELETED;
+  }
+
+  // Handle situation where new user exists already with data.
+  if (newUser && hasUserData(newUser)) {
+    return LinkAccountResult.NEW_USER_HAS_DATA;
+  }
+
+  // New user exists but has no data (linking should continue as if new user does not exist).
```
zognin: When will this case happen? I think it's a bit confusing to think about it as `oldUser` & `newUser` compared to `anonUser` and `verifiedUser` because the only time we should `linkAccount` is between an `anonUser` and `verifiedUser`?

Once `verifiedUser` exists, any future attempts to link an `anonUser` with `verifiedUser` shouldn't work?

types/user.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Body of the POST request to '/api/user/linkAccount'.
+ *
+ * The 'old_token' field indicates the JWT token of the old account that needs to be replaced by the new account.
+ * The details of the new (current) account will be provided by the Authorization header (Bearer type) in the request
+ * as per normal.
+ */
+export interface AccountPostData {
```
zognin, line 8: Maybe we can remove this if it's not used

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+
+async function linkAccount(oldUid: string, newUid: string): Promise<LinkAccountResult> {
+  const newUser = await getUserIfExists(newUid);
+  const oldUser = await getUserIfExists(oldUid);
+
+  // If old user has already been removed from the database. This can happen if the API was called twice or if the new user
+  // creation process failed and the API was retried.
+  if (!oldUser) {
+    if (!newUser) {
+      await prisma.user.create({ data: { uid: newUid } });
```
zognin: Besides, is this correct now because the `oldUser` isn't deleted?

pages/api/user/index.ts
sivayogasubramanian: Let's standardize to use Object.freeze format

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  applications: { id: number }[];
+}
+
+enum LinkAccountResult {
+  OLD_USER_DELETED,
+  NEW_USER_HAS_DATA,
+  NEW_USER_LINKED,
+}
+
+const { ERROR, INFORMATION, SUCCESS } = StatusMessageType;
```
sivayogasubramanian: Lets not do this as me and gn are not doing this. Good to standardize.

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+    case HttpMethod.DELETE:
+      return handleDelete(currentUid, req, res);
+    default:
+      return rejectHttpMethod(res, req.method);
+  }
+}
+
+async function handlePostWithoutOldToken(
+  currentUid: string,
+  _req: NextApiRequest,
```
sivayogasubramanian: Let's not pass in req if it is not used.

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+        .json(
+          createJsonResponse(
+            {},
+            { type: SUCCESS, message: `Old UID ${oldUid} has been replaced with new UID ${currentUid}.` },
+          ),
+        );
+  }
+}
+
+async function handleDelete(uid: string, _req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
```
sivayogasubramanian: Same here for req

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  // UID is a primary key, so the count will either be 0 (UID does not exist in database) or 1 (successful delete).
+  const user = await getUserIfExists(uid);
+  if (!user) {
+    return res
+      .status(HttpStatus.NOT_FOUND)
+      .json(createJsonResponse({}, { type: ERROR, message: `Could not find UID ${uid}.` }));
+  }
+
+  await prisma.application.deleteMany({ where: { userId: uid } });
```
sivayogasubramanian: If I am not wrong, applications will be deleted when the user is deleted due to the `onDelete: Cascade` constraint?

pages/api/user/index.ts
sivayogasubramanian: For all the `StatusMessages` in this file, is it possible to refactor it to make it user facing? I agree with @zognin that UID should not be exposed.

pages/api/user/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  NEW_USER_HAS_DATA,
+  NEW_USER_LINKED,
+}
+
+const { ERROR, INFORMATION, SUCCESS } = StatusMessageType;
+const prisma = new PrismaClient();
+
+async function handler(currentUid: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
+  switch (req.method) {
+    case HttpMethod.POST:
```
sivayogasubramanian, line 60: Lets have a `handlePost` method and move the logic in the switch statement there.

pages/api/user/index.ts
sivayogasubramanian: Also lets send over the user as a payload

pages/api/user/index.ts
sivayogasubramanian: Yes. I am also confused about this. During our discussion, I thought we agreed on just updating the anon id -> verified id when user logs in with github. Why would there be a need to delete from the db?

pages/api/user/index.ts
bharathcs: What does sending over the user as a payload entail? Just the user itself or the deeply nested applications and stages / tasks?

sivayogasubramanian: Just the user.

pages/api/user/index.ts
bharathcs: if anonUid is updated to verifiedUid, the original anonUid is no longer in the database. I said "already removed from database" to indicate at the point of the request, the old uid might have been deleted or already linked to some other, unknown verifiedUid.

pages/api/user/index.ts
bharathcs: This could happen in the following flow:

1. User opens on browser X (anon1)
2. User signs in with GitHub on X (verified1): there is still no data
3. User opens up on browser Y (anon2)
4. User adds data on browser Y (anon2)
5. User remembers about the GitHub accounts.
6. User signs in with Github on Y (verified1)

With my flow, on step 6 verified1 gets the data created by anon2.

We could deal with it the way you mentioned, but I wanted to improve UX and not reject an attempt to link if the verified doesn't have data.

pages/api/user/index.ts
bharathcs: Not sure what you mean here, sorry.

pages/api/user/index.ts
bharathcs: My mistake, I thought it didn't.

pages/api/user/index.ts
bharathcs: Is this a nit or is it important to you? Cuz I don't see how that is clearer, extracting these few lines into another function makes things harder to follow IMO. A bit of logic in the switch case makes the different handler flows obvious.

bharathcs:
Oops, deleted instead of resolve.

pages/api/user/index.ts
sivayogasubramanian: Its a nit. This method seemed weird because the abstraction level is different for this switch case (according to SLAP at least).

pages/api/user/index.ts
zognin: I don't think this makes sense because once a `verifiedUser` is created, we cannot assume that future log ins of an `anonUser` to a `verifiedUser` means that `anonUser` is the `verifiedUser`.

Actually even the first time is a bit iffy isn't it? Should we never take anon user's data when we log in?

Suppose you're tracking your data anonymously on someone else's browser but you forget to clear cache, then anyone else can log in and take your data?

pages/api/user/index.ts
zognin: When we link account by your method of creating a new user, why is the old user not deleted?

pages/api/user/index.ts
bharathcs: Not possible to 'take' someone else's data. If you forgot to clear cache, there's a log in button. To log in with GitHub, you would be clicking something like 'Link account', similar to other web apps out there. If you can reach the 'link account' while logged in (aka with the anon token still in ur browser), then you get that data.

pages/api/user/index.ts
bharathcs: Ah I get that. I will remove this flow, it doesn't make sense that old user is not deleted.

pages/api/user/index.ts
bharathcs: Resolved: we will never allow relinking.

pages/api/user/index.ts
bharathcs: Resolved: we will never allow relinking.

pages/api/user/index.ts
bharathcs: Hmm fair..

pages/api/user/index.ts
bharathcs: Moved to its own issue: #29

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+async function handlePostWithoutOldToken(currentUid: string, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
+  const isExisting = (await getUserIfExists(currentUid)) !== null;
+  if (isExisting) {
+    return res.status(HttpStatus.CONFLICT).json(createJsonResponse({}, messages[MessageType.USER_ALREADY_EXISTS]));
+  }
+
+  await prisma.user.create({ data: { uid: currentUid } });
+  return res.status(HttpStatus.CREATED).json(createJsonResponse({}, messages[MessageType.USER_CREATED]));
```
sivayogasubramanian: Return uid here too?

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  USER_CREATED,
+  USER_ALREADY_EXISTS,
+  // DELETE accounts
+  USER_DELETED,
+  DELETE_TARGET_NOT_FOUND,
+}
+
+const messages = Object.freeze({
+  // POST messages (without oldToken)
+  [MessageType.USER_CREATED]: { type: StatusMessageType.SUCCESS, message: 'New user with UID added.' },
```
sivayogasubramanian:
```suggestion
  [MessageType.USER_CREATED]: { type: StatusMessageType.SUCCESS, message: 'New user added.' },
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  [MessageType.USER_ALREADY_EXISTS]: { type: StatusMessageType.ERROR, message: 'User with UID already exists.' },
```
sivayogasubramanian:
```suggestion
  [MessageType.USER_ALREADY_EXISTS]: { type: StatusMessageType.ERROR, message: 'User already exists.' },
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  // POST messages (with oldToken)
+  [MessageType.NEW_USER_LINKED]: {
+    type: StatusMessageType.SUCCESS,
+    message: 'Old UID has been replaced with new UID.',
```
sivayogasubramanian:
```suggestion
    message: 'Account was linked successfully.',
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  },
+  [MessageType.NEW_USER_UNVERIFIED]: {
+    type: StatusMessageType.ERROR,
+    message: 'An unverified account cannot be the target of link.',
```
sivayogasubramanian:
```suggestion
    message: 'Account was not linked.',
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  },
+  [MessageType.INVALID_OLD_TOKEN]: {
+    type: StatusMessageType.ERROR,
+    message: 'Could not validate and decode "oldToken".',
```
sivayogasubramanian:
```suggestion
    message: 'There was an error while retrieving user info.',
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  let oldUserDetails: UserDetailsFromRequest;
+  try {
+    oldUserDetails = await getUserFromJwt(oldUserToken);
+  } catch (error) {
+    return res.status(HttpStatus.UNAUTHORIZED).json(createJsonResponse({}, messages[MessageType.INVALID_OLD_TOKEN]));
+  }
+
+  const oldUid = oldUserDetails.uid;
+  if (!oldUserDetails.isAnonymous) {
```
sivayogasubramanian, line 108: After the discussion, we allow linking 1 time right? Will this case ever happen?

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+    message: 'An unverified account cannot be the target of link.',
+  },
+  [MessageType.INVALID_OLD_TOKEN]: {
+    type: StatusMessageType.ERROR,
+    message: 'Could not validate and decode "oldToken".',
+  },
+  [MessageType.OLD_USER_VERIFIED]: { type: StatusMessageType.ERROR, message: 'Old UID is already linked.' },
+  [MessageType.NEW_USER_EXISTS]: {
+    type: StatusMessageType.ERROR,
+    message: 'New UID already exists and cannot be linked.',
```
sivayogasubramanian:
```suggestion
    message: 'There is already a user associated with this GitHub account.',
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  },
+  [MessageType.OLD_USER_NOT_FOUND]: { type: StatusMessageType.ERROR, message: 'Old UID not found.' },
```
sivayogasubramanian:
```suggestion
  [MessageType.OLD_USER_NOT_FOUND]: { type: StatusMessageType.ERROR, message: 'There was an error while retrieving user info.' },
```

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  // DELETE messages.
+  [MessageType.USER_DELETED]: { type: StatusMessageType.SUCCESS, message: 'Deleted UID and their data.' },
```
sivayogasubramanian: We no longer need this.

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  [MessageType.DELETE_TARGET_NOT_FOUND]: { type: StatusMessageType.ERROR, message: 'Could not find UID to delete.' },
```
sivayogasubramanian: Same here.

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  OLD_USER_VERIFIED,
+  // POST account creation only.
+  USER_CREATED,
+  USER_ALREADY_EXISTS,
+  // DELETE accounts
+  USER_DELETED,
+  DELETE_TARGET_NOT_FOUND,
+}
+
+const messages = Object.freeze({
```
sivayogasubramanian, line 33: I made a lot of changes to the messages to make it user facing. Feel free to veto and resolve if you want.

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  // Handle situation where new user exists already with data.
+  if (newUser) {
+    return MessageType.NEW_USER_EXISTS;
+  }
+
+  // New user exists but has no data (linking should continue as if new user does not exist).
+  if (newUser) {
+    await prisma.user.delete({ where: { uid: newUid } });
+  }
```
sivayogasubramanian: This if statement will never execute.

pages/api/user/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  // DELETE messages.
+  [MessageType.USER_DELETED]: { type: StatusMessageType.SUCCESS, message: 'Deleted UID and their data.' },
+  [MessageType.DELETE_TARGET_NOT_FOUND]: { type: StatusMessageType.ERROR, message: 'Could not find UID to delete.' },
+});
+
+async function handler(currentUid: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<UserData>>) {
+  switch (req.method) {
+    case HttpMethod.POST:
+      return handlePost(currentUid, req, res);
+    case HttpMethod.DELETE:
```
sivayogasubramanian, line 62: Lets delete this DELETE endpoint for how. I don't see a use case for this now.

pages/api/user/index.ts
bharathcs: This isn't about relinking *to* an existing verified account but rather trying to link *from* an existing verified. Just defensive programming to make sure no one tries to "Link with Github" twice somehow.

pages/api/user/index.ts
bharathcs: Resolved: we are keeping DELETE

pages/api/user/index.ts
bharathcs: Resolved: we are keeping DELETE

pages/api/user/index.ts
bharathcs: Thanks!

pages/api/user/index.ts
bharathcs: Resolved: we are keeping DELETE

pages/api/user/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      return handleDelete(currentUid, res);
+    default:
+      return rejectHttpMethod(res, req.method);
+  }
+}
+
+function handlePost(currentUid: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<UserData>>) {
+  const { oldToken } = req.body as AccountPostData;
+  const oldUserToken = oldToken && typeof oldToken == 'string' ? oldToken.trim() : oldToken;
```
ganhongyao:
```suggestion
  const oldUserToken = oldToken && typeof oldToken === 'string' ? oldToken.trim() : oldToken;
```

Nit

pages/api/user/index.ts
ganhongyao: Should this line be removed then?

ganhongyao (approved):

ganhongyao:
LGTM, thanks! Can merge once the above 2 points are addressed.

PR #22, opened 2022-09-09
Add stages api logic.

bharathcs:


pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+import { createJsonResponse, HttpMethod, HttpStatus, rejectHttpMethod } from '../../../../../utils/http/httpHelpers';
+import { withPrismaErrorHandling } from '../../../../../utils/prisma/prismaHelpers';
+import { ApplicationStageData, ApplicationStagePostData } from '../../../../../types/applicationStage';
+import { isValidDate } from '../../../../../utils/date/validations';
+import { withAuthUser } from '../../../../../utils/auth/jwtHelpers';
+import { Nullable } from '../../../../../types/utils';
 
-export default function handler(req: NextApiRequest, res: NextApiResponse) {
-  res.status(200).end(`Request sent to ${req.url}\n`);
+const { ERROR, SUCCESS } = StatusMessageType;
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Same as last time.

types/applicationStage.ts
```diff
@@ -11,6 +11,11 @@ export type ApplicationStageApplicationData = {
 
 export type ApplicationStageData = ApplicationStageApplicationData & { applicationId: number };
 
+export type ApplicationStagePostData = {
```
sivayogasubramanian, line 14: Why is there no emoji and remark here?

They should be here but optional right?

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  INVALID_APPLICATION_ID,
+  APPLICATION_NOT_FOUND,
+  CREATED,
+}
+
+const messages = Object.freeze({
+  [MessageType.INVALID_TYPE]: {
+    type: ERROR,
+    message:
+      'Application Stage type is invalid. Type must be one of: ' + Object.values(ApplicationStageType).join(', ') + '.',
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
      'Application stage type is invalid.',
```

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  },
+  [MessageType.INVALID_DATE]: {
+    type: ERROR,
+    message: 'Application Stage date is invalid.',
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
    message: 'Application stage date is invalid.',
```

nit

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  },
+  [MessageType.INVALID_APPLICATION_ID]: {
+    type: ERROR,
+    message: 'Query parameter applicationId (.../applications/:applicationId/...) is invalid.',
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Is it possible to make this more user facing?

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  },
+  [MessageType.APPLICATION_NOT_FOUND]: {
+    type: ERROR,
+    message: 'User does not have any application with the given ID.',
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Let's standardise to messages that are used for this message type in other files.

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  }
+
+  const application = await getApplication(uid, Number(req.query.applicationId));
+  if (application === null) {
+    return res.status(HttpStatus.NOT_FOUND).json(createJsonResponse({}, messages[MessageType.APPLICATION_NOT_FOUND]));
+  }
+
+  switch (req.method) {
+    case HttpMethod.POST:
+      return postHandler(application, req, res);
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
      return handlePost(application, req, res);
```
nit: for standardisation

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+    case HttpMethod.GET:
+      return getHandler(application, res);
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
      return handleGet(application, res);
```

nit: for standardisation. also can move this above?

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+function validatePathParameters(req: NextApiRequest): Nullable<MessageType> {
+  const applicationId = req.query.applicationId as string;
+  return Number.isInteger(applicationId) ? MessageType.INVALID_APPLICATION_ID : null;
+}
+
+/** Validates path parameters and body. */
+function validatePostRequest(req: NextApiRequest): Nullable<MessageType> {
+  const { type, date } = req.body;
+
+  if (typeof type != 'string' || !(type in ApplicationStageType)) {
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Is using `!=` here better than `!==`. I always that we always use `!==` for strict equality.

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+async function getHandler(
+  application: ApplicationDataWithStages,
+  res: NextApiResponse<ApiResponse<ApplicationDataWithStages>>,
+) {
+  return res.status(HttpStatus.OK).json(createJsonResponse(application, messages[MessageType.CREATED]));
+}
+
+/** Validates path parameters and body. */
+function validatePathParameters(req: NextApiRequest): Nullable<MessageType> {
+  const applicationId = req.query.applicationId as string;
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: You don't need this line.

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  return Number.isInteger(applicationId) ? MessageType.INVALID_APPLICATION_ID : null;
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
  return !Number.isInteger(req.query.applicationId) ? MessageType.INVALID_APPLICATION_ID : null;
```

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+}
+
+/** Validates path parameters and body. */
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: I think this comment causes more confusion than it helps. Where does it validate the params?

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
 }
+
+async function getHandler(
+  application: ApplicationDataWithStages,
+  res: NextApiResponse<ApiResponse<ApplicationDataWithStages>>,
+) {
+  return res.status(HttpStatus.OK).json(createJsonResponse(application, messages[MessageType.CREATED]));
+}
+
+/** Validates path parameters and body. */
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Same here. I think the method name is clear.

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  }
+
+  const { type, date } = req.body;
+  const parameters: ApplicationStagePostData = { type, date: new Date(date) };
+
+  const newStage: ApplicationStageData = await prisma.applicationStage.create({
+    data: { ...parameters, applicationId: application.id },
+    select: { applicationId: true, id: true, type: true, date: true, emojiUnicodeHex: true, remark: true },
+  });
+  return res.status(HttpStatus.CREATED).json(createJsonResponse(newStage, messages[MessageType.CREATED]));
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Nit: add a line break before this.

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+import { isValidDate } from '../../../../../utils/date/validations';
+import { withAuthUser } from '../../../../../utils/auth/jwtHelpers';
+import { Nullable } from '../../../../../types/utils';
 
-export default function handler(req: NextApiRequest, res: NextApiResponse) {
-  res.status(200).end(`Request sent to ${req.url}\n`);
+const { ERROR, SUCCESS } = StatusMessageType;
+const prisma = new PrismaClient();
+
+type ApplicationDataWithStages = Application & { applicationStages: ApplicationStage[] };
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Why is this here? should we consolidate all types in the types dir?

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,133 @@
...
+  const parameters: ApplicationStagePostData = { type, date: new Date(date) };
+
+  const newStage: ApplicationStageData = await prisma.applicationStage.create({
+    data: { ...parameters, applicationId: application.id },
+    select: { applicationId: true, id: true, type: true, date: true, emojiUnicodeHex: true, remark: true },
+  });
+  return res.status(HttpStatus.CREATED).json(createJsonResponse(newStage, messages[MessageType.CREATED]));
 }
+
+async function getHandler(
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: Why is the GET /api/applications/:id/stages returning the application?

Should this be done in GET /api/applications/:id instead?

Comment (sivayogasubramanian) [reply pages/api/applications/[applicationId]/stages/index.ts]: We also have no need for this route.

types/applicationStage.ts
bharathcs: We're only creating a new stage here, not adding reactions right? Doesn't make sense to combine the two actions I felt.

1. User creates stage
2. User adds reaction

If the UI is definitely gonna treat it as two actions, make sense to keep it as such w backend.

Same as why POST tasks doesn't support specifying isDone tbh.

pages/api/applications/[applicationId]/stages/index.ts
Comment (bharathcs) [reply pages/api/applications/[applicationId]/stages/index.ts]: I wasn't really concerned because typeof should always return a string, a non-strict equals will work the same way but I can fix this for standardisation. 

Could we add this to the es lint rules too?

pages/api/applications/[applicationId]/stages/index.ts
Comment (bharathcs) [reply pages/api/applications/[applicationId]/stages/index.ts]: Oof thanks for the catch.

pages/api/applications/[applicationId]/stages/index.ts
Comment (bharathcs) [reply pages/api/applications/[applicationId]/stages/index.ts]: Got it. I'm also going to change `/pages/api/applications/[applicationId].ts` `APPLICATION_DOES_NOT_BELONG_TO_USER` to match `APPLICATION_NOT_FOUND` because we should not reveal to a user whether the application exists for someone else.

pages/api/applications/[applicationId]/stages/index.ts
Comment (bharathcs) [reply pages/api/applications/[applicationId]/stages/index.ts]: Ah.. yes 😞

types/applicationStage.ts
sivayogasubramanian: @zognin @ganhongyao wdyt?

bharathcs:
Feel free to ignore the test file: they're not meant to be good quality or wtv, just to reduce effort for whoever wants to use them.

__tests__/pages/api/applications/[applicationId]/stages/index.test.ts
```diff
@@ -0,0 +1,148 @@
...
+beforeAll(prismaMigrateReset);
+afterAll(prismaMigrateReset);
+
+const createStagePostMocks = (uid: string, applicationId: number, body: ApplicationStagePostData) =>
+  createMocks(stagesHandler, { uid, method: HttpMethod.POST, body, query: { applicationId: applicationId + '' } });
+
+describe('POST to create stage works.', () => {
+  const prisma = getPrismaClientForTests();
+
+  it('Allows for POST creation', async () => {
```
Comment (bharathcs) [inline __tests__/pages/api/applications/[applicationId]/stages/index.test.ts]:
```suggestion
  it('can create with POST', async () => {
```

bharathcs (commented):
Trigger rerun.

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,146 @@
...
+    (typeof emojiUnicodeHex === 'string' && isValidHex(emojiUnicodeHex));
+  if (!isValidEmojiUnicodeHex) {
+    return MessageType.INVALID_EMOJI_UNICODE_HEX;
+  }
+
+  return null;
+}
+
+/** Finds application if present for the user, else returns null. */
+async function getApplication(uid: string, applicationId: number): Promise<Nullable<ApplicationDataWithStagesOnly>> {
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]: We are only checking if the application is found.

I don't think we need this method and the type.

types/application.ts
```diff
@@ -12,6 +13,8 @@ export type ApplicationData = {
   tasks: TaskData[];
 };
 
+export type ApplicationDataWithStagesOnly = Application & { applicationStages: ApplicationStage[] };
```
sivayogasubramanian: I think this type will not be needed.

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,146 @@
...
+  }
+}
+
+async function handlePost(uid: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<ApplicationStageData>>) {
+  const paramValidationError = validatePostRequest(req);
+  if (paramValidationError !== null) {
+    return res.status(HttpStatus.BAD_REQUEST).json(createJsonResponse({}, messages[paramValidationError]));
+  }
+
+  const application = await getApplication(uid, Number(req.query.applicationId));
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
  const application = await prisma.application.findFirst({
    where: { id: Number(req.query.applicationId), userId: uid },
  });
```

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,146 @@
...
+  if (application === null) {
+    return res.status(HttpStatus.NOT_FOUND).json(createJsonResponse({}, messages[MessageType.APPLICATION_NOT_FOUND]));
+  }
+
+  const { type, date, emojiUnicodeHex, remark } = req.body;
+  // Empty or undefined emojis and remarks will be set to null.
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
  
```

pages/api/applications/[applicationId]/stages/index.ts
```diff
@@ -1,5 +1,146 @@
...
+  const parameters: ApplicationStagePostData = {
+    type,
+    date: new Date(date),
+    emojiUnicodeHex: emojiUnicodeHex || null,
+    remark: remark || null,
```
Comment (sivayogasubramanian) [inline pages/api/applications/[applicationId]/stages/index.ts]:
```suggestion
    emojiUnicodeHex,
    remark,
```

I don't think we need this. I tested at it works without.

sivayogasubramanian (approved):
LGTM! Thanks

PR #23, opened 2022-09-09
Add logic for .../tasks/:taskId/.

bharathcs:


pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  INVALID_NOTIFICATION_DATETIME,
+  INVALID_IS_DONE,
+}
+
+const messages = Object.freeze({
+  [MessageType.INVALID_TASK_ID]: { type: StatusMessageType.ERROR, message: 'Task id is invalid.' },
+  [MessageType.INVALID_APPLICATION_ID]: { type: StatusMessageType.ERROR, message: 'Application id is invalid.' },
+  [MessageType.TASK_NOT_FOUND]: { type: StatusMessageType.ERROR, message: 'Task cannot be found.' },
+  [MessageType.TASK_DELETED]: { type: StatusMessageType.SUCCESS, message: 'The task has been successfully deleted.' },
+  [MessageType.TASK_PATCHED]: { type: StatusMessageType.SUCCESS, message: 'The task has been successfully patched.' },
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: Maybe we can use updated because users may not know what `patched` means.
```suggestion
  [MessageType.TASK_UPDATED_SUCCESSFULLY]: { type: StatusMessageType.SUCCESS, message: 'The task has been successfully updated.' },
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  INVALID_DUE_DATE,
+  INVALID_NOTIFICATION_DATETIME,
+  INVALID_IS_DONE,
+}
+
+const messages = Object.freeze({
+  [MessageType.INVALID_TASK_ID]: { type: StatusMessageType.ERROR, message: 'Task id is invalid.' },
+  [MessageType.INVALID_APPLICATION_ID]: { type: StatusMessageType.ERROR, message: 'Application id is invalid.' },
+  [MessageType.TASK_NOT_FOUND]: { type: StatusMessageType.ERROR, message: 'Task cannot be found.' },
+  [MessageType.TASK_DELETED]: { type: StatusMessageType.SUCCESS, message: 'The task has been successfully deleted.' },
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: For consistency with other apis
```suggestion
  [MessageType.TASK_DELETED_SUCCESSFULLY]: { type: StatusMessageType.SUCCESS, message: 'The task has been successfully deleted.' },
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  [MessageType.INVALID_DUE_DATE]: { type: StatusMessageType.ERROR, message: 'The task due date is invalid.' },
+  [MessageType.INVALID_NOTIFICATION_DATETIME]: {
+    type: StatusMessageType.ERROR,
+    message: 'The task notification date is invalid.',
+  },
+  [MessageType.INVALID_IS_DONE]: { type: StatusMessageType.ERROR, message: 'The task completion status is invalid.' },
+});
+
+async function handler(uid: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<TaskData>>) {
+  const { taskId, applicationId } = req.query as { taskId: string; applicationId: string };
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
  const { taskId, applicationId } = req.query as TaskRouteParams;
```

Should we define the type of the params in `types/user.ts`?

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+
+  if (!Number.isInteger(Number(taskId))) {
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: I noticed that we are repeatedly doing `Number(taskId)` and `Number(applicationId)`. Should we declare these as variables?

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+      .json(createJsonResponse({}, messages[MessageType.INVALID_APPLICATION_ID]));
+  }
+
+  const task = await prisma.task.findUnique({
+    where: { id: Number(taskId) },
+    include: { application: { select: { id: true, userId: true } } },
+  });
+  if (!task || task.application.id !== Number(applicationId) || task.application.userId !== uid) {
+    return res.status(HttpStatus.NOT_FOUND).json(createJsonResponse({}, messages[MessageType.TASK_NOT_FOUND]));
+  }
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: Do we want to use this level of abstraction here? I guess it reduces code duplication in some ways, but it also breaks the current level of abstraction and increases coupling?

Not sure if it's right to put the responsibility of checking `applicationId` or uniqueness or that the task belongs to the user in the top level handler too because what if other methods/apis route to this endpoint in future and don't have the same constraints?

These checks are implicitly handled here so when I look at the `deleteHandler` for example, I started to wonder if we already checked that the user is authorized to delete the task.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+
+  switch (req.method) {
+    case HttpMethod.PATCH:
+      return patchHandler(task.id, req, res);
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: For consistency with other files. Otherwise we can rename everything if `XHandler` seems better? Same for `deleteHandler`
```suggestion
      return handlePatch(task.id, req, res);
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: Do you think this part can be extracted out into a function, and then let `patchHandler` and `deleteHandler` call this method instead? I think it is better for the `handler` to just remain stupid and hand off the request to the appropriate method handler, and also to be consistent with the other routes currently.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+    case HttpMethod.DELETE:
+      return deleteHandler(task.id, req, res);
+    default:
+      rejectHttpMethod(res, req.method);
+  }
+}
+
+async function patchHandler(taskId: number, req: NextApiRequest, res: NextApiResponse<ApiResponse<TaskData>>) {
+  const validationError = validateTaskPatchData(req);
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
  const errorMessageType = validatePatchRequest(req);
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+
+async function deleteHandler(taskId: number, _req: NextApiRequest, res: NextApiResponse<ApiResponse<TaskData>>) {
+  await prisma.task.delete({ where: { id: taskId } });
+  return res.status(HttpStatus.OK).json(createJsonResponse({}, messages[MessageType.TASK_DELETED]));
+}
+
+function validateTaskPatchData(req: NextApiRequest): MessageType | null {
+  const { title, dueDate, notificationDateTime, isDone } = req.body;
+
+  const isValidTitle = title === undefined || (typeof title === 'string' && title?.trim().length === 0);
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
  const isValidTitle = title === undefined || (typeof title === 'string' && title?.trim().length > 0);
```
Do you mean this instead? A title is only valid if it is not empty

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  }
+}
+
+async function patchHandler(taskId: number, req: NextApiRequest, res: NextApiResponse<ApiResponse<TaskData>>) {
+  const validationError = validateTaskPatchData(req);
+  if (validationError !== null) {
+    return res.status(HttpStatus.BAD_REQUEST).json(createJsonResponse({}, messages[validationError]));
+  }
+
+  const data: TaskPatchData = req.body;
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: https://github.com/sivayogasubramanian/caza/pull/15#discussion_r967629027. But I know the downside is that we must manually ensure to only assign defined fields.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  const updatedTask: Task = await prisma.task.update({ where: { id: taskId }, data });
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts:79]: Do we want to check `where: { ..., applicationId: applicationId}` too? It's not strictly necessary but can seem a bit if the user puts in random application ids into the path.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  return res.status(HttpStatus.OK).json(createJsonResponse(updatedTask, messages[MessageType.TASK_PATCHED]));
 }
+
+async function deleteHandler(taskId: number, _req: NextApiRequest, res: NextApiResponse<ApiResponse<TaskData>>) {
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
async function deleteHandler(taskId: number, _req: NextApiRequest, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  await prisma.task.delete({ where: { id: taskId } });
+  return res.status(HttpStatus.OK).json(createJsonResponse({}, messages[MessageType.TASK_DELETED]));
+}
+
+function validateTaskPatchData(req: NextApiRequest): MessageType | null {
+  const { title, dueDate, notificationDateTime, isDone } = req.body;
+
+  const isValidTitle = title === undefined || (typeof title === 'string' && title?.trim().length === 0);
+  const isValidDueDate = dueDate === undefined || !isValidDate(dueDate + '');
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
  const isValidDueDate = dueDate === undefined || (typeof dueDate === 'string' && isValidDate(dueDate));
```

Condition flipped here as well, and I think we should only accept `dueDate` which are `strings`.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  const isValidNotificationDateTime =
+    notificationDateTime === undefined ||
+    notificationDateTime === null ||
+    (typeof notificationDateTime === 'string' && !isValidDate(notificationDateTime));
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
    (typeof notificationDateTime === 'string' && isValidDate(notificationDateTime));
```
Condition flipped here as well

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,116 @@
...
+  const updatedTask: Task = await prisma.task.update({ where: { id: taskId }, data });
+  return res.status(HttpStatus.OK).json(createJsonResponse(updatedTask, messages[MessageType.TASK_PATCHED]));
 }
+
+async function deleteHandler(taskId: number, _req: NextApiRequest, res: NextApiResponse<ApiResponse<TaskData>>) {
+  await prisma.task.delete({ where: { id: taskId } });
+  return res.status(HttpStatus.OK).json(createJsonResponse({}, messages[MessageType.TASK_DELETED]));
+}
+
+function validateTaskPatchData(req: NextApiRequest): MessageType | null {
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
function validateTaskPatchData(req: NextApiRequest): Nullable<MessageType> {
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,114 @@
...
+
+async function deleteHandler(taskId: number, res: NextApiResponse<ApiResponse<EmptyPayload>>) {
+  await prisma.task.delete({ where: { id: taskId } });
+  return res.status(HttpStatus.OK).json(createJsonResponse({}, messages[MessageType.TASK_DELETED]));
+}
+
+function validateTaskPatchData(req: NextApiRequest): MessageType | null {
+  const { title, dueDate, notificationDateTime, isDone } = req.body;
+
+  const isValidTitle = title === undefined || (typeof title === 'string' && title?.trim().length !== 0);
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: Maybe we can use the validation function that we have?
```suggestion
  const isValidTitle = title === undefined || (typeof title === 'string' && !isEmpty(title));
```

bharathcs:
Feel free to ignore the __test__ file: they're not meant to be good quality or wtv, just to reduce effort for whoever wants to use them.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,114 @@
...
+  const isValidDueDate = dueDate === undefined || isValidDate('' + dueDate);
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
  const isValidDueDate = dueDate === undefined || (typeof dueDate === 'string' && isValidDate(dueDate));
```

Should we only accept `dueDate` as valid if it is of type `string`? If not passing a number will pass the validation too.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,114 @@
...
+  const isValidNotificationDateTime =
+    notificationDateTime === undefined || notificationDateTime === null || isValidDate('' + notificationDateTime);
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
    notificationDateTime === undefined || notificationDateTime === null || (typeof notificationDateTime === 'string' && isValidDate(notificationDateTime));
```
Same as above

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,114 @@
...
+  const isValidDoneParameter = isDone === undefined || typeof isDone === 'boolean';
+
+  if (!isValidTitle) {
+    return MessageType.INVALID_TITLE;
+  }
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts:125]: Do you think it'll be good if we arrange this method to return the errors right after the check? So that the flow is more logical.

Something like 

```typescript
  ...
  const isValidTitle = title === undefined || (typeof title === 'string' && title?.trim().length === 0);
  if (!isValidTitle) {
    return MessageType.INVALID_TITLE;
  }

  const isValidDueDate = dueDate === undefined || typeof dueDate === 'string' && !isValidDate(dueDate);
  if (!isValidDueDate) {
    return MessageType.INVALID_DUE_DATE;
  }

  ... more validations ...
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
Comment (ganhongyao) [reply pages/api/applications/[applicationId]/tasks/[taskId].ts]: Resolving outdated comment

zognin (commented):
Haven't tested yet, here are some first thoughts

pages/api/applications/[applicationId]/tasks/[taskId].ts
Comment (sivayogasubramanian) [reply pages/api/applications/[applicationId]/tasks/[taskId].ts]: Why was this resolved without any changes? I think we either change everything to `XHandler` (including other files) or change this to `handlePatch`

Comment (bharathcs) [reply pages/api/applications/[applicationId]/tasks/[taskId].ts]: 🤦  my bad.

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,163 @@
...
+  const isValidDoneParameter = isDone === undefined || typeof isDone === 'boolean';
+  if (!isValidDoneParameter) {
+    return MessageType.INVALID_IS_DONE;
+  }
+
+  return null;
+}
+
+async function getTask(uid: string, req: NextApiRequest) {
+  const { applicationId, taskId } = req.query as unknown as TaskQueryParams;
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]: Do we need `as unknown`?
```suggestion
  const { applicationId, taskId } = req.query as TaskQueryParams;
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,163 @@
...
+  }
+  const taskId = taskOrError.id;
+
+  await prisma.task.delete({ where: { id: taskId } });
+  return res.status(HttpStatus.OK).json(createJsonResponse({}, messages[MessageType.TASK_DELETED_SUCCESSFULLY]));
+}
+
+function validatePathParameters(req: NextApiRequest): Nullable<MessageType> {
+  const { taskId, applicationId } = req.query;
+  if (!taskId && !Number.isInteger(Number(taskId))) {
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
  if (!canBecomeInteger(taskId)) {
```

pages/api/applications/[applicationId]/tasks/[taskId].ts
```diff
@@ -1,5 +1,163 @@
...
+    return MessageType.INVALID_TASK_ID;
+  }
+
+  if (!Number.isInteger(Number(applicationId))) {
```
Comment (zognin) [inline pages/api/applications/[applicationId]/tasks/[taskId].ts]:
```suggestion
  if (!canBecomeInteger(applicationId)) {
```

zognin (approved):
Pre-approving. LGTM.

PR #24, opened 2022-09-10
Add Jest for API tests.

bharathcs:
Only use this if you feel it saves you time!

- Modify hello handler to match the rest.
- Add jest module.
- Add hello.test file.
- Complicated set up required for simple jest test writing.


sivayogasubramanian (approved):
LGTM!

PR #26, opened 2022-09-10
Fix bugs with prev apis

### LINKED ISSUE #21: During create, `title` is a compulsory field.
During create, `title` is a compulsory field.
This is different for patch, or for creating entities with non-compulsory fields.
You may want to check for the other fields as well.

```suggestion
 if (!req.body.title) {
   // Error that title is missing
 }

  if (isEmpty(req.body.title)) {
```
Btw, not including `req.body.title` leads to runtime errors. I'm not sure if `withPrismaErrorHandling` was supposed to catch that.

_Originally posted by @zognin in https://github.com/sivayogasubramanian/caza/pull/12#discussion_r967010369_

sivayogasubramanian:
Fixes #21

### DIFF
--- a/pages/api/applications/[applicationId]/tasks/index.ts
+++ b/pages/api/applications/[applicationId]/tasks/index.ts
@@ -14,6 +14,7 @@ const prisma = new PrismaClient();
 
 enum MessageType {
   MISSING_TITLE,
+  INVALID_TITLE,
   EMPTY_TITLE,
   MISSING_DUE_DATE,
   INVALID_DUE_DATE,
@@ -26,6 +27,7 @@ enum MessageType {
 
 const messages = Object.freeze({
   [MessageType.MISSING_TITLE]: { type: StatusMessageType.ERROR, message: 'Task title is missing.' },
+  [MessageType.INVALID_TITLE]: { type: StatusMessageType.ERROR, message: 'Task title is invalid.' },
   [MessageType.EMPTY_TITLE]: { type: StatusMessageType.ERROR, message: 'Task title is empty.' },
   [MessageType.MISSING_DUE_DATE]: {
     type: StatusMessageType.ERROR,
@@ -115,6 +117,10 @@ function validateRequest(req: NextApiRequest): Nullable<MessageType> {
     return MessageType.MISSING_TITLE;
   }
 
+  if (req.body.title === null) {
+    return MessageType.INVALID_TITLE;
+  }
+
   if (isEmpty(req.body.title)) {
     return MessageType.EMPTY_TITLE;
   }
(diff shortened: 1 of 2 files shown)


pages/api/roles/index.ts
```diff
@@ -104,24 +117,48 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<
...
+  if (req.body.companyId === null || typeof req.body.companyId !== 'number') {
     return MessageType.INVALID_COMPANY_ID;
   }
 
-  if (req.body.title && isEmpty(req.body.title)) {
+  if (req.body.title === undefined) {
+    return MessageType.MISSING_TITLE;
+  }
+
+  if (req.body.title === null) {
```
```suggestion
  if (req.body.title === null || typeof req.body.companyId !== 'string') {
```
If not we get error when we try to do `.trim()` in line 136 as part of `isEmpty`

pages/api/roles/index.ts
```diff
@@ -104,24 +117,48 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<
...
+  if (isEmpty(req.body.title)) {
     return MessageType.EMPTY_TITLE;
   }
 
-  if ((req.body.type && isEmpty(req.body.type)) || !(req.body.type in RoleType)) {
+  if (req.body.type === undefined) {
+    return MessageType.MISSING_ROLE_TYPE;
+  }
+
+  if (req.body.type === null || isEmpty(req.body.type) || !(req.body.type in RoleType)) {
```
```suggestion
  if (req.body.type === null || typeof req.body.companyId !== 'string' || isEmpty(req.body.type) || !(req.body.type in RoleType)) {
```

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -104,16 +113,28 @@ async function handlePost(userId: string, req: NextApiRequest, res: NextApiRespo
...
-  if (req.body.title && isEmpty(req.body.title)) {
+  if (req.body.title === undefined) {
+    return MessageType.MISSING_TITLE;
+  }
+
+  if (req.body.title === null) {
+    return MessageType.INVALID_TITLE;
+  }
+
+  if (isEmpty(req.body.title)) {
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/index.ts:124]: We should check that it is a string, same as above comments

pages/api/applications/[applicationId]/tasks/index.ts
```diff
@@ -104,16 +113,28 @@ async function handlePost(userId: string, req: NextApiRequest, res: NextApiRespo
...
+  }
+
+  if (req.body.dueDate === null || !isValidDate(req.body.dueDate)) {
     return MessageType.INVALID_DUE_DATE;
   }
 
   if (
-    (req.body.notification || req.body.notificationDateTime !== undefined) &&
+    (req.body.notification === null || req.body.notificationDateTime !== undefined) &&
     !isValidDate(req.body.notificationDateTime)
```
Comment (ganhongyao) [inline pages/api/applications/[applicationId]/tasks/index.ts]: Are we doing the `isValidDate` check only when the date is not null and not undefined? In this case should it be like that? 

 ```suggestion
     (req.body.notification !== null && req.body.notificationDateTime !== undefined) &&
     !isValidDate(req.body.notificationDateTime)
```


I don't mind doing this either, it checks for both `null` and `undefined`.

 ```suggestion
     req.body.notification != null &&
     !isValidDate(req.body.notificationDateTime))
```

ganhongyao (approved):

PR #27, opened 2022-09-10
Complete roles api

### LINKED ISSUE #25: Complete roles API

sivayogasubramanian:
Fixes #25

### DIFF
--- a/pages/api/roles/index.ts
+++ b/pages/api/roles/index.ts
@@ -72,7 +72,7 @@ async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<R
 
   const queryCompanyId = typeof companyId == 'string' && isInteger(companyId) ? Number(companyId) : undefined;
 
-  const searchWords = typeof searchQuery == 'string' && !isEmpty(searchQuery) ? searchQuery.split(/\s+/) : [];
+  const searchWords = typeof searchQuery == 'string' && !isEmpty(searchQuery) ? searchQuery.trim().split(/\s+/) : [];
 
   const searchInts = searchWords
     .filter(isInteger)
--- a/pages/api/roles/index.ts
+++ b/pages/api/roles/index.ts
@@ -1,4 +1,4 @@
-import { PrismaClient, RoleType } from '@prisma/client';
+import { PrismaClient, Role, RoleType } from '@prisma/client';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { ApiResponse, StatusMessageType } from '../../../types/apiResponse';
 import { RoleData, RoleListData, RolePostData } from '../../../types/role';
@@ -6,8 +6,9 @@ import { Nullable } from '../../../types/utils';
 import { withAuth } from '../../../utils/auth/jwtHelpers';
 import { MIN_ROLE_YEAR } from '../../../utils/constants';
 import { createJsonResponse, HttpMethod, HttpStatus, rejectHttpMethod } from '../../../utils/http/httpHelpers';
+import { isInteger } from '../../../utils/numbers/validations';
 import { withPrismaErrorHandling } from '../../../utils/prisma/prismaHelpers';
-import { capitalizeEveryWord } from '../../../utils/strings/formatters';
+import { capitalizeEveryWord, trim } from '../../../utils/strings/formatters';
 import { isEmpty } from '../../../utils/strings/validations';
 
 const prisma = new PrismaClient();
@@ -66,12 +67,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
   }
 }
 
-// TODO: Optional Search query. Will be done in another PR to unblock others.
-
 async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
+  const { companyId, title, type, year } = req.query;
+
+  const queryCompanyId = isInteger(companyId as string) ? Number(companyId) : undefined;
+  const queryTitle = title ? trim(title) : undefined;
+  const queryType = type && (type as string) in RoleType ? (trim(type) as RoleType) : undefined;
+  const queryYear = isInteger(year as string) ? Number(year) : undefined;
+
+  if (!queryCompanyId && !queryTitle && !queryType && !queryYear) {
+    res.status(HttpStatus.OK).json(createJsonResponse([]));
+    return;
+  }
+
   const roles: RoleListData[] = await prisma.role.findMany({
     where: {
       isVerified: true,
+      companyId: queryCompanyId,
+      title: { contains: queryTitle },
+      type: queryType,
+      year: queryYear,
     },
     select: {
       id: true,
--- a/pages/api/roles/index.ts
+++ b/pages/api/roles/index.ts
@@ -8,7 +8,7 @@ import { MIN_ROLE_YEAR } from '../../../utils/constants';
 import { createJsonResponse, HttpMethod, HttpStatus, rejectHttpMethod } from '../../../utils/http/httpHelpers';
 import { isInteger } from '../../../utils/numbers/validations';
 import { withPrismaErrorHandling } from '../../../utils/prisma/prismaHelpers';
-import { capitalizeEveryWord, trim } from '../../../utils/strings/formatters';
+import { capitalizeEveryWord } from '../../../utils/strings/formatters';
 import { isEmpty } from '../../../utils/strings/validations';
 
 const prisma = new PrismaClient();
@@ -82,9 +82,7 @@ async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<R
       },
     }));
 
-  console.log(queryWords);
-
-  if (!queryCompanyId) {
+  if (!queryCompanyId || !searchWords.length) {
     res.status(HttpStatus.OK).json(createJsonResponse([]));
     return;
   }
--- a/pages/api/roles/index.ts
+++ b/pages/api/roles/index.ts
@@ -1,7 +1,7 @@
-import { PrismaClient, RoleType } from '@prisma/client';
+import { Prisma, PrismaClient, RoleType } from '@prisma/client';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { ApiResponse, StatusMessageType } from '../../../types/apiResponse';
-import { RoleData, RoleListData, RolePostData } from '../../../types/role';
+import { RoleData, RoleListData, RolePostData, RoleQueryParams } from '../../../types/role';
 import { Nullable } from '../../../types/utils';
 import { withAuth } from '../../../utils/auth/jwtHelpers';
 import { MIN_ROLE_YEAR } from '../../../utils/constants';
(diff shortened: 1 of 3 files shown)


pages/api/roles/index.ts
```diff
@@ -68,12 +69,32 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
 }
 
-// TODO: Optional Search query. Will be done in another PR to unblock others.
-
 async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
+  const { companyId, searchQuery } = req.query;
+
+  const queryCompanyId = isInteger(companyId as string) ? Number(companyId) : undefined;
+
+  const searchWords = (searchQuery as string) ? (searchQuery as string).split(' ') : [];
```
bharathcs: Do we know for sure that searchQuery is a string? I'm getting this error when I pass a number:
```error - TypeError: 1.split is not a function```

```suggestion
  const searchWords = typeof searchQuery === "string" ? (searchQuery as string).split(' ') : [];
```

pages/api/roles/index.ts
```diff
@@ -68,12 +69,32 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  const queryYear = searchWords.filter(isInteger).map(Number)[0];
```
bharathcs: A bit far fetched but a internship from 5 Guys / Forever 21 could set year with the wrong number (e.g. `Forever 21 Summer Intern 2022`)
```suggestion
  const queryYear = searchWords.filter(isInteger).map(Number).filter(x => x >= MIN_YEAR)[0];
```

pages/api/roles/index.ts
```diff
@@ -68,12 +69,35 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
 }
 
-// TODO: Optional Search query. Will be done in another PR to unblock others.
-
 async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
+  const { companyId, searchQuery } = req.query;
+
+  const queryCompanyId = isInteger(companyId as string) ? Number(companyId) : undefined;
+
+  const searchWords = typeof searchQuery === 'string' ? (searchQuery as string).split(' ') : [];
```
ganhongyao: I feel like these type assertions are getting out of hand and are littered around the codebase, do you think this is a better approach? 

```suggestion
type QueryParams = {
  companyId: number;
  searchQuery: string;
}

async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
  const { companyId, searchQuery } = req.query as QueryParams;

  const queryCompanyId = isInteger(companyId) ? Number(companyId) : undefined;

  const searchWords = (searchQuery).split(' ');
```

```suggestion
  const searchWords = !isEmpty(searchQuery) ? searchQuery.split(/\s+/) : [];
```

So that we split by any amount of whitespace. Also we should set it to an empty array if the search query is an empty string. If not we will get `contains: ''` filters.

pages/api/roles/index.ts
```diff
@@ -68,12 +69,35 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  const queryYear = searchWords
+    .filter(isInteger)
+    .map(Number)
+    .filter((year) => year >= MIN_ROLE_YEAR)[0];
```
ganhongyao: Is it safe to assume that the filtered array will always be non-empty?

pages/api/roles/index.ts
```diff
@@ -68,12 +69,35 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  const queryWords = searchWords
+    .filter((word) => !(isInteger(word) && Number(word) >= MIN_ROLE_YEAR))
+    .map((word) => ({
+      title: {
+        contains: word,
+      },
+    }));
+
+  if (!queryCompanyId || !searchWords.length) {
```
ganhongyao: Not returning any results when no company is specified is fine, but we should still show results when there are no search words.

pages/api/roles/index.ts
```diff
@@ -68,12 +69,35 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  const searchWords = typeof searchQuery === 'string' ? (searchQuery as string).split(' ') : [];
+  const queryYear = searchWords
+    .filter(isInteger)
+    .map(Number)
+    .filter((year) => year >= MIN_ROLE_YEAR)[0];
+  const queryWords = searchWords
+    .filter((word) => !(isInteger(word) && Number(word) >= MIN_ROLE_YEAR))
+    .map((word) => ({
+      title: {
+        contains: word,
```
ganhongyao, line 87:
```suggestion
        contains: word,
        mode: 'insensitive',
```

For case-insensitive matches

pages/api/roles/index.ts
```diff
@@ -68,12 +69,35 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  if (!queryCompanyId || !searchWords.length) {
+    res.status(HttpStatus.OK).json(createJsonResponse([]));
+    return;
+  }
+
   const roles: RoleListData[] = await prisma.role.findMany({
     where: {
       isVerified: true,
+      companyId: queryCompanyId,
+      AND: [{ year: { equals: queryYear } }],
```
```suggestion
      AND: queryYear ? [{ year: { equals: queryYear } }] : [],
```

Since we cannot assume that there will always be a query year

pages/api/roles/index.ts
```diff
@@ -68,12 +69,35 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
...
+      OR: [...queryWords],
```
```suggestion
      OR: queryWords.length > 0 ? queryWords : undefined,
```

> OR
One or more conditions must return true.

In case there are no query words (e.g. `searchString` is `2023`, we should set it to undefined. If not, we will never get any matches.

This is the query generated when we pass an empty array to `OR`:

```sql
prisma:query SELECT "public"."roles"."id", "public"."roles"."title", "public"."roles"."type", 
"public"."roles"."year", "public"."roles"."is_verified", "public"."roles"."company_id" FROM "public"."roles" 
WHERE("public"."roles"."is_verified" = $1 AND "public"."roles"."company_id" = $2 AND 
"public"."roles"."year" = $3 AND 1=0) OFFSET $4 /* traceparent=00-00-00-00 */
```

Note the 1=0

pages/api/roles/index.ts
ganhongyao: As for checking the type of `searchQuery` to make sure that it is of type `string` and not `string[]`, I think that we can leave it out, since we can assume that we are the ones specifying the query params from the frontend and we will always pass a string. I don't mind if you add it back in though.

ganhongyao:
One more thing: I assume with the `companyId` filter, this is going to be purely for autocomplete, and not for listing all roles? If so, we should add a `take` option to `findMany` to limit our number of results returned (to maybe 5 or something).

pages/api/roles/index.ts
bharathcs: We will still need type assertions, but using a query interface looks good to me. But I would prefer `QueryParams` interface to be in the `/types` instead of in the `/pages/api/{handler file}.ts`. 

There already is `{api name}{api method}Data` (e.g. `RolePostData`) interface in the `/types/roles.ts` path. 
If we also put the `{api name}{api method}QueryParams` (e.g. `RolePostQueryParams`) in the `/types/roles.ts` path it will be useful for future client side API work.

bharathcs: Wait actually how do you know `searchQuery` is a string?

bharathcs: Would still prefer to keep the typeof since we don't know that it is a string.
```suggestion
async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
  const { companyId, searchQuery } = req.query;

  const queryCompanyId = isInteger(companyId as string) ? Number(companyId) : undefined;

  const searchWords = typeof searchQuery === 'string' ? (searchQuery as string).split(' ').filter(x => !isEmpty(x) : [];
```

pages/api/roles/index.ts
bharathcs: nice catch.

pages/api/roles/index.ts
```diff
@@ -68,12 +69,35 @@ function handler(req: NextApiRequest, res: NextApiResponse) {
   }
 }
 
-// TODO: Optional Search query. Will be done in another PR to unblock others.
-
 async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
```
bharathcs, line 70: Also, could you merge #35 to fix the handlers in this PR (and elsewhere)

pages/api/roles/index.ts
bharathcs: And also since companyId is already casted as a number, `isInteger(companyId)` won't compile I think...

```suggestion
function validateQueryParams(req.query) {
  const { companyId, searchQuery } = req.query;
  if (!isInteger(companyId as string)) return undefined;
  
  return { 
    companyId: Number(companyId), 
    searchQuery: typeof searchQuery === 'string' ? searchQuery : "";
  }
}

async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<RoleListData[]>>) {
  const queryParams = validateQueryParams(req.query);
  if (!queryParams) { .. reject 400 }

  const { companyId, searchQuery }  = queryParams;
  const searchWords = typeof searchQuery === 'string' ? (searchQuery as string).split(' ') : [];
```
Hmm idk if this is any better...

pages/api/roles/index.ts
sivayogasubramanian: Thanks!

pages/api/roles/index.ts
ganhongyao:
```
type QueryParams = {
  companyId: number;
  searchQuery: string;
}
```
@bharathcs This does not work, and if you try to do this Typescript will complain that these types are not overlapping, since the type defined for `query` is 
```
    query: Partial<{
        [key: string]: string | string[];
    }>;
```
and `number` and `string` do not overlap.

> Wait actually how do you know searchQuery is a string?

We don't, although we can implicitly assume that it will be since we are the ones specifying the query parameters from the frontend. But to be safe we can keep the type-checking `typeof searchQuery === 'string'`. (And I just realised that with these explicit typechecks we don't need the type assertions anymore which means my original comment is irrelevant, but I guess we can still keep `RoleQueryParams`.)

ganhongyao (dismissed):
LGTM

pages/api/roles/index.ts
```diff
@@ -66,13 +67,41 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  const { companyId, searchQuery } = req.query as RoleQueryParams;
+
+  const queryCompanyId = typeof companyId == 'string' && isInteger(companyId) ? Number(companyId) : undefined;
+
+  const searchWords = typeof searchQuery == 'string' && !isEmpty(searchQuery) ? searchQuery.split(/\s+/) : [];
```
```suggestion
  const searchWords = typeof searchQuery == 'string' && !isEmpty(searchQuery) ? searchQuery.trim().split(/\s+/) : [];
```

So that we don't get empty strings at the start or end after splitting.

ganhongyao (approved):

PR #28, opened 2022-09-10
Ensure that middleware doesn't break promise contract.

bharathcs:
There was an issue earlier where middleware wasn't returning promises that were chained after the handler. This could result in serious bugs down the road (during serverless), but its fixed now.


sivayogasubramanian (commented):
LGTM!

sivayogasubramanian (approved):

PR #30, opened 2022-09-10
Fix early return for handler.

bharathcs:
This handler returns before the async promise was completed.


sivayogasubramanian:
@zognin resolved this in an earlier PR

PR #32, opened 2022-09-10
Add basic logic for API calls

ganhongyao:
Changes
- Add `swr` which is used for caching and sending axios fetch requests
   > SWR is a strategy to first return the data from cache (stale), then send the fetch request (revalidate), and finally come with the up-to-date data.
- Add axios for api calls 
- Add sample `/roles` route to test fetching of roles and creation of roles.


api/interceptor.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+      config.headers.Authorization = `Bearer ${jwtToken}`;
+      return config;
+    };
+
+    const interceptor = api.interceptors.request.use(requestInterceptor);
+    setIsIntercepted(true);
+
+    return () => api.interceptors.request.eject(interceptor);
+  }, [jwtToken]);
```
bharathcs:
```suggestion
  }, [jwtToken, setIsIntercepted]);
```

api/api.ts
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios';
+
+const api = axios.create({
+  baseURL: '/api/',
```
bharathcs, line 4: In case our server doesn't respond

```suggestion
  baseURL: '/api/',
  timeout: 10 * 1000, // 10 seconds.
```

bharathcs (approved):
Nice work!

api/rolesApi.ts
```diff
@@ -0,0 +1,17 @@
...
+  private getRolesUrl() {
+    return 'roles';
+  }
+
+  public createRole(role: RolePostData): ApiPromise<RoleData> {
+    return api.post(this.getRolesUrl(), role).then((res) => res.data);
+  }
+}
+
+const rolesApi = new RolesApi();
```
bharathcs: Btw Hong Yao, is there a reason we want to use classes? I feel the functional way has less boiler plate and easier to write.

```suggestion
const ROLES_URL = 'roles';

export function createRole(role: RolePostData): ApiPromise<RoleData> {
   return api.post(ROLES_URL, role).then((res) => res.data);
}

// export function getRoles(companyId: number, searchQuery?: string): ApiPromise<RoleData> { ... }
// export function deleteRole..
```

```
rolesApi.createRole(...)
```

ganhongyao: I feel that having an additional namespace makes it clear that this is an API call. There may be more similarly named functions down the road such as Redux actions and functions within your components. What do you think about this?

bharathcs: Makes sense. Can we keep the roles url as a const instead of a function then?

components/forms/CreateRoleForm.tsx
```diff
@@ -0,0 +1,50 @@
...
+const { Option } = Select;
+
+function CreateRoleForm() {
+  const [form] = Form.useForm();
+
+  const onSubmit = (values: RolePostData) => {
+    rolesApi.createRole(values).then(() => {
+      form.resetFields();
+      alert('Role created successfully!');
+    });
```
bharathcs:
```suggestion
    }).catch((error) => alert(error)); // temporarily while we set up error displays.
```

bharathcs (approved):
Feel free to resolve the comments from my previous review if you like

bharathcs (approved):
LGTM!

pages/roles/index.tsx
```diff
@@ -0,0 +1,32 @@
+import useSWR from 'swr';
+import { Spin, Table, Typography } from 'antd';
+import CreateRoleForm from '../../components/forms/CreateRoleForm';
+import { ROLES_URL } from '../../api/rolesApi';
+
+const { Title } = Typography;
+
+function Applications() {
+  const { data } = useSWR(ROLES_URL);
```
bharathcs: Can we have it as ROLES_API_ENDPOINT to be more descriptive? Super nit only, feel free to ignore.

bharathcs (approved):

PR #36, opened 2022-09-11
Add functions and error types for validating stage chronology.

bharathcs:
This will be needed for POST /applications/:applicationId/stages and PATCH /applications/:applicationId/stages/:stageId

Cuz I realised we should have backend validation for the stages being in the right order.


bharathcs:
Affects @zognin 's #15 (merged) and my #22 (WIP) if we want to support this.

types/applicationStage.ts
```diff
@@ -17,3 +17,60 @@ export type ApplicationStagePatchData = {
...
+export function validateStageChronology(
+  ...stages: { type: ApplicationStageType; date: Date }[]
+): Nullable<ChronologicalValidationError> {
+  const finalStages = stages.filter((stage) => FINAL_STAGES.includes(stage.type));
+  if (finalStages.length > 1) {
+    return ChronologicalValidationError.MORE_THAN_ONE_FINAL_STAGE;
+  }
+
+  if (finalStages.length !== 0) {
+    const finalStage = stages[0];
```
ganhongyao:
```suggestion
    const finalStage = finalStages[0];
```

Should it be this instead?

types/applicationStage.ts
```diff
@@ -17,3 +17,60 @@ export type ApplicationStagePatchData = {
...
+    const firstStage = firstStages[0];
+    for (const stage of stages) {
+      if (stage.date < firstStage.date) {
+        return ChronologicalValidationError.FIRST_STAGE_OUT_OF_ORDER;
+      }
+    }
+  }
+
+  return null;
+}
```
ganhongyao: Do you think it would be better to split this function into smaller functions, each validating for just one condition?

```typescript
// inside validateStageChronology
...
const hasMultipleFinalStages = hasMoreThanOneFinalStage(stages); // returns a boolean
if (hasMultipleFinalStages) {
  return ChronologicalValidationError.MORE_THAN_ONE_FINAL_STAGE;
}
...
```

types/applicationStage.ts
```diff
@@ -17,3 +17,60 @@ export type ApplicationStagePatchData = {
...
+ * Takes in a list of stages and if any of the following conditions are out of order, returns a message.
+ * - final stage (withdrawn / accepted / rejected) occurs before another stage
+ * - more than one final stage
+ * - first stage (applied) occurs after another stage
+ * - more than one first stage
+ *
+ * In the interest of user flexibility, having zero first / final stages is tolerated, as is having final / first
+ * stages on the same date as other stages.
+ */
+export function validateStageChronology(
```
ganhongyao: Should this be in another file and keep only type declarations in our `types` files?

ganhongyao (approved):
LGTM 😊

PR #38, opened 2022-09-11
Fix validation of inputs that can be an integer

zognin:
- Fields that should be `number` should still use `Number.isInteger`
- Fields that can be any type, but can become an integer, should use `canBeInteger`
- Note: `Number('')`, `Number(false)` and `Number(null)` gives 0 in javascript. Edit: Will guard against these.

### DIFF
--- a/utils/numbers/validations.ts
+++ b/utils/numbers/validations.ts
@@ -1,0 +1,4 @@
+export function canBecomeInteger(value: unknown) {
+  const number = Number(value);
+  return !isNaN(number) && Number.isInteger(number);
+}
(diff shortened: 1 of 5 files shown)


utils/numbers/validations.ts
```diff
@@ -0,0 +1,4 @@
+export function canBecomeInteger(value: unknown) {
+  const number = Number(value);
+  return !isNaN(number) && Number.isInteger(number);
```
bharathcs: Actually "falsey" values like 0, "", and undefined return 0 when passed to `Number()`

```suggestion
  return (!value && value !== 0) && !isNaN(number) && Number.isInteger(number);
```

bharathcs: Sorry my bad! Didn't read your description. I am in favour of guarding against these cases within this function.

bharathcs (approved):
LGTM!

PR #39, opened 2022-09-12
Use null instead of setting to empty string.

bharathcs:
Right now setting to empty string is valid.

For simplicity can we have setting to empty string be the same as deleting contents (null).


zognin (approved):
LGTM thanks

PR #40, opened 2022-09-12
Add applications search

zognin:
This is the search logic:
Return applications that match {any of role years} AND {any of role types} AND {any role title or company name that matches keyword} AND {latestStage.type matches any of stage types}

Also refactored roles search. @sivayogasubramanian please check and test if the logic is still correct.

### DIFF
--- a/pages/api/applications/index.ts
+++ b/pages/api/applications/index.ts
@@ -25,8 +25,6 @@ enum MessageType {
   DUPLICATE_APPLICATION,
   MISSING_ROLE_ID,
   INVALID_ROLE_ID,
-  INVALID_QUERY_APPLICATION_STAGE_TYPE,
-  INVALID_QUERY_ROLE_TYPE,
 }
 
 const prisma = new PrismaClient();
@@ -42,14 +40,6 @@ const messages = Object.freeze({
   },
   [MessageType.MISSING_ROLE_ID]: { type: StatusMessageType.ERROR, message: 'Application role id is missing.' },
   [MessageType.INVALID_ROLE_ID]: { type: StatusMessageType.ERROR, message: 'Application role id is invalid.' },
-  [MessageType.INVALID_QUERY_ROLE_TYPE]: {
-    type: StatusMessageType.ERROR,
-    message: 'Application query role type is invalid.',
-  },
-  [MessageType.INVALID_QUERY_APPLICATION_STAGE_TYPE]: {
-    type: StatusMessageType.ERROR,
-    message: 'Application query stage type is invalid.',
-  },
 });
 
 async function handler(userId: string, req: NextApiRequest, res: NextApiResponse) {
@@ -70,17 +60,11 @@ async function handleGet(
   res: NextApiResponse<ApiResponse<ApplicationListData[]>>,
 ) {
   const queryParams: ApplicationQueryParams = parseGetQueryParams(req);
-  const errorMessageType = validateGetQueryParams(queryParams);
-  if (errorMessageType !== null) {
-    res.status(HttpStatus.BAD_REQUEST).json(createJsonResponse({}, messages[errorMessageType]));
-    return;
-  }
 
   const companyNameFilters = makeCompanyNameFilters(queryParams.searchWords);
   const roleTitleFilters = makeRoleTitleFilters(queryParams.searchWords);
   const roleYearFilters = makeRoleYearFilters(queryParams.searchWords);
-  // Safe to typecast as validation is done above.
-  const roleTypeFilters = makeRoleTypeFilters(queryParams.roleTypeWords as RoleType[]);
+  const roleTypeFilters = makeRoleTypeFilters(queryParams.roleTypeWords);
   const companyOrFilters = companyNameFilters?.map((filter) => ({ company: filter }));
   const roleTitleOrCompanyFilters = getArrayOrUndefined<Prisma.RoleWhereInput>(
     combineDefinedArrays<Prisma.RoleWhereInput>([roleTitleFilters, companyOrFilters]),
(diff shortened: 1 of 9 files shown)


utils/date/validations.ts
```diff
@@ -1,3 +1,14 @@
+import { MIN_ROLE_YEAR } from '../constants';
```
sivayogasubramanian: Let's rename this to MIN_YEAR due to the naming of the functions here.

utils/filters/filterHelpers.ts
```diff
@@ -0,0 +1,55 @@
...
+  // Filtering for year is done separately.
+  const roleTitleSearchWords = searchWords
+    .filter((word) => !canBecomeValidYear(word))
+    .map((word) => ({
+      title: {
+        contains: word,
+        mode: Prisma.QueryMode.insensitive,
+      },
+    }));
+  return roleTitleSearchWords.length === 0 ? undefined : roleTitleSearchWords;
```
sivayogasubramanian: Is it better to use getNonEmptyArrayOrUndefined here?

utils/filters/filterHelpers.ts
```diff
@@ -0,0 +1,55 @@
...
+  const roleYearSearchWords = searchWords
+    .filter(canBecomeInteger)
+    .map(Number)
+    .filter(isValidYear)
+    .map((yearKeyword) => ({
+      year: {
+        equals: yearKeyword,
+      },
+    }));
+  return roleYearSearchWords.length === 0 ? undefined : roleYearSearchWords;
```
sivayogasubramanian: Same here.

pages/api/roles/index.ts
```diff
@@ -68,38 +69,16 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  const roleYearFilters = makeRoleYearFilters(queryParams.searchWords);
 
   const roles: RoleListData[] = await prisma.role.findMany({
     where: {
       isVerified: true,
-      companyId: queryCompanyId,
-      AND: queryYear ? [{ year: { equals: queryYear } }] : undefined,
-      OR: queryWords.length > 0 ? queryWords : undefined,
+      companyId: queryParams.companyId,
+      AND: roleYearFilters,
```
sivayogasubramanian:
```suggestion
      OR: roleYearFilters,
```

Would a OR search make more sense here? If i type '2023 2024', i will get no results according to this logic.

pages/api/roles/index.ts
```diff
@@ -143,6 +122,14 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<
   res.status(HttpStatus.CREATED).json(createJsonResponse(newRole, messages[MessageType.ROLE_CREATED_SUCCESSFULLY]));
 }
 
+function parseGetQueryParams(req: NextApiRequest): RoleQueryParams {
+  const { companyId, searchQuery } = req.query;
+  const searchWords =
+    searchQuery === undefined ? [] : Array.isArray(searchQuery) ? searchQuery : searchQuery.trim().split(/\s+/);
```
sivayogasubramanian: Are we matching word by word? 

What if searchWords is an array like ["software engineer", "site re"]?

utils/filters/filterHelpers.ts
```diff
@@ -0,0 +1,55 @@
+import { Prisma, RoleType } from '@prisma/client';
+import { canBecomeValidYear, isValidYear } from '../date/validations';
+import { canBecomeInteger } from '../numbers/validations';
+
+export function makeCompanyNameFilters(searchWords: string[]) {
+  // Do not want to return names that contain the year to prevent zero results.
```
sivayogasubramanian, line 6: I don't understand this. Is it possible to explain?

utils/filters/filterHelpers.ts
```diff
@@ -0,0 +1,55 @@
...
+            contains: word,
+            mode: Prisma.QueryMode.insensitive,
+          },
+        }));
+}
+
+export function makeRoleTitleFilters(searchWords: string[]) {
+  // Do not want to return titles that contain the year.
+  // This prevents unrelated titles that only match the year but not the other search words.
+  // Filtering for year is done separately.
```
sivayogasubramanian: What does this mean?

utils/filters/filterHelpers.ts
```diff
@@ -0,0 +1,55 @@
...
+      },
+    }));
+  return roleTitleSearchWords.length === 0 ? undefined : roleTitleSearchWords;
+}
+
+export function makeRoleYearFilters(searchWords: string[]) {
+  const roleYearSearchWords = searchWords
+    .filter(canBecomeInteger)
+    .map(Number)
+    .filter(isValidYear)
```
sivayogasubramanian:
```suggestion
    
```

utils/filters/filterHelpers.ts
```diff
@@ -0,0 +1,55 @@
...
+        contains: word,
+        mode: Prisma.QueryMode.insensitive,
+      },
+    }));
+  return roleTitleSearchWords.length === 0 ? undefined : roleTitleSearchWords;
+}
+
+export function makeRoleYearFilters(searchWords: string[]) {
+  const roleYearSearchWords = searchWords
+    .filter(canBecomeInteger)
```
sivayogasubramanian:
```suggestion
    .filter(canBecomeValidYear)
```

pages/api/applications/index.ts
```diff
@@ -159,6 +194,39 @@ async function handlePost(userId: string, req: NextApiRequest, res: NextApiRespo
     .json(createJsonResponse(newApplication, messages[MessageType.APPLICATION_CREATED_SUCCESSFULLY]));
 }
 
+function parseGetQueryParams(req: NextApiRequest): ApplicationQueryParams {
+  const searchWords =
+    req.query.searchQuery === undefined
+      ? []
+      : Array.isArray(req.query.searchQuery)
+      ? req.query.searchQuery
```
sivayogasubramanian: Same as the previous comment.

pages/api/applications/index.ts
```diff
@@ -159,6 +194,39 @@ async function handlePost(userId: string, req: NextApiRequest, res: NextApiRespo
...
+      : req.query.searchQuery.trim().split(/\s+/);
+
+  const roleTypeUncheckedWords =
+    req.query.roleTypes === undefined
+      ? []
+      : Array.isArray(req.query.roleTypes)
+      ? req.query.roleTypes
```
sivayogasubramanian: Same as the previous comment.

pages/api/applications/index.ts
```diff
@@ -159,6 +194,39 @@ async function handlePost(userId: string, req: NextApiRequest, res: NextApiRespo
...
+      : req.query.roleTypes.trim().split(/\s*,\s*/);
+
+  // Safe to typecast due to the filter check.
+  const roleTypeWords: RoleType[] = roleTypeUncheckedWords.filter((word) => word in RoleType) as RoleType[];
+
+  const stageTypeUncheckedWords =
+    req.query.stageTypes === undefined
+      ? []
+      : Array.isArray(req.query.stageTypes)
+      ? req.query.stageTypes
```
sivayogasubramanian: Same as the previous comment.

pages/api/roles/index.ts
zognin: For our use search words is supposed to be a string and each item in the array should be considered individual units to compare against if we want to support other cases in future.

I think it's not worth to split words in the array for now, although we can do it.

sivayogasubramanian: Lets save time by not splitting the words in the array for now.

utils/filters/filterHelpers.ts
zognin: Based on the logic in the description, if we do something like `?searchQuery=2023`, this will return empty payload if there are no company names with `2023`

The alternative logic that I tried is too loose and returns too many unrelated applications.

utils/filters/filterHelpers.ts
zognin: If we find titles that match `2023`, suppose we search `cloud 2023`, we still get all other titles with `2023` like `software engineering intern 2023` which is too wide.

I will change the comment.

sivayogasubramanian (approved):
LGTM! Thanks for refactoring my parts 😀.

PR #43, opened 2022-09-13
Implement frontend APIs

sivayogasubramanian:
NOTE: Styling for the toasts will not be done in this PR.


utils/url.ts
```diff
@@ -0,0 +1,30 @@
...
+      return toQueryString(value, key);
+    } else if (value === undefined || value === null) {
+      return '';
+    } else {
+      return `${key}=${encodeURIComponent(value)}`;
+    }
+  });
+
+  return query.filter((str) => str.length > 0).join('&');
+}
```
bharathcs, line 30: Hey Siva, is there a reason we can't use the axios way of handling query parameters?

```
await api.get('path', { params }).then(...);
```

params supports all the normal objects expected in a GET query, and I feel its safer to trust the axios implementation.

pages/api/companies.ts
```diff
@@ -111,6 +110,13 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<
     .json(createJsonResponse(newCompany, messages[MessageType.COMPANY_CREATED_SUCCESSFULLY]));
 }
 
+function parseGetQueryParams(req: NextApiRequest): CompanyQueryParams {
+  const { companyNames } = req.query;
+  const names =
+    companyNames === undefined ? [] : Array.isArray(companyNames) ? companyNames : splitByWhitespaces(companyNames);
+  return { companyNames: names };
```
bharathcs, line 117: I think its better to require the companyName to be of type `string[] | undefined`, rather than allowing for split by whitespace. We'd have to then consider issues like "National University of Singapore" treated as 4 different company names.

utils/url.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Converts an object into a URI query string.
+ * Supports arrays and nested objects.
+ */
+// eslint-disable-next-line
+export function toQueryString(object: Record<string, any>, prefix = ''): string {
```
bharathcs, line 6: Hey Siva, is there a reason we can't use the axios way of handling query parameters?

```
await api.get('path', { params }).then(...);
```

params supports all the normal objects expected in a GET query, and will also serialises it in the JSON way expected by servers. I feel its safer to trust the axios implementation.

api/applicationStages.ts
```diff
@@ -0,0 +1,40 @@
+import { ApiPromise, EmptyPayload } from '../types/apiResponse';
```
bharathcs: Can we rename the file to be *Api.ts to be consistent with the other files in this folder?

bharathcs (commented):
Thanks Siva, it looks good. I haven't tested (the query param thing) yet but I just had a few comments to point out.

Another thing: I see we're doing the `.then(res => res.data)` everywhere, but we don't have a standardised way to handle errors. I personally am in favour of requiring callers of `/api/*Api.apiFunction()` to handle the errors in a consistent way. Since we expect messages to be used how does this sound:


```typescript
export function makeErrorsConsistent<D>(apiPromise: AxiosPromise<ApiResponse<D>>): Promise<ApiResponse<D>> {
  return apiPromise
    .then((res) => res.data)
    .catch((error: AxiosError) => {
      if (!error.request) {
        return { messages: [{ type: StatusMessageType.ERROR, message: 'Unable to complete the requested action.' }] };
      } else if (!error.response) {
        return { messages: [{ type: StatusMessageType.ERROR, message: 'No response from the server.' }] };
      }
      const { status, statusText, data } = error.response as AxiosResponse<ApiResponse<D> | undefined>;
      if (data && data.messages) {
        return { messages };
      }
      return { messages: [{ type: StatusMessageType.ERROR, message: `Unable to complete request: ${statusText} (${status})` }] };
    });
}
```

This way every API caller will know they have to handle the specified result. If you prefer throwing errors instead, we can modify `makeErrorsConsistent` to throw a custom `ApiError` that presents these messages. 

```typescript
class ApiError extends Error {
  public messages: StatusMessage[];
  constructor(messages) {
    super("API promise was rejected.");
    this.messages = messages;
  }
}
```

utils/url.ts
bharathcs: If we want to do `toQueryString()` anyway for our own custom path parameter serialization, we should move this function to the axios api interceptor file.


```typescript
interceptors.request.use((config) => {
  config.paramsSerializer = /* goes here */
  return config
})
```

sivayogasubramanian: I'll go with the interceptor suggestion. This is because when we use the axios way, the query params for array look like this `searchWords[]=soft` instead of `searchWords=soft` which is what next expects.

pages/api/companies.ts
sivayogasubramanian: image

Its considered as one. I don't get what you mean here.

api/applicationStagesApi.ts
```diff
@@ -0,0 +1,34 @@
+import { ApiPromise, EmptyPayload } from '../types/apiResponse';
+import { ApplicationStageData, ApplicationStagePatchData, ApplicationStagePostData } from '../types/applicationStage';
+import api from './api';
+import { APPLICATIONS_API_ENDPOINT } from './applicationsApi';
+
+export const APPLICATION_STAGES_API_ENDPOINT = 'stages';
+
+class ApplicationStagesApi {
+  public createApplicationStage(
```
bharathcs, line 9: Sorry I didn't consider this before, but what about just using 'create()', 'edit()' etc? Since the class name already is clear, I think the function can be more succint.

```typescript
applicationStageApi.create() // vs
applicationStageApi.createApplicationStage()
```

api/api.ts
```diff
@@ -1,8 +1,90 @@
...
+    };
+  }
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  return error.response.data;
+}
+
+function getResponseMessages(response: ApiResponse<any>): string {
+  return response.messages?.length > 0 ? response.messages.map((m) => m.message).join(' : ') : '-';
```
bharathcs, line 81: Messages are supposed to be used as toasts right? I feel like concatenating by ' : ' or returning '-' is quite user unfriendly.

If there are no messages, there shouldn't be a toast (return null?). If there are more than one messages (e.g. `[ 'Could not find X.', 'Type Y is invalid.', 'Type G is invalid.' ]` I would rather concatenate all without a separator so it reads like 3 sentences to the user.

```suggestion
  return response.messages?.map((m) => m.message).join(' ') || null;
```

bharathcs: If you do want to keep the current implementation, minor nit:

```suggestion
  return response.messages?.map((m) => m.message).join(' : ') || '-';
```

api/api.ts
```diff
@@ -1,8 +1,90 @@
...
+
+      return apiResponse;
+    })
+    .catch((error: AxiosError) => {
+      if (process.env.NODE_ENV === 'development') {
+        console.error(`[API] ${error.code} ${endpoint} : ${error.message}`);
+      }
+
+      throw makeApiErrorResponse(error);
+    });
```
bharathcs, line 61: Minor nit: suppress typescript async suggestion:
image

```suggestion
    }) as ApiPromise<D>;
```

For posterity: we can't convert to async function as we're using `ApiPromise`, a type alias for `Promise<...>`. [Known issue](https://github.com/microsoft/TypeScript/issues/26781) with typescript compiler.

api/api.ts
```diff
@@ -1,8 +1,90 @@
...
+}
+
+function makeApiErrorResponse(error: AxiosError): ApiResponse<EmptyPayload> {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  if (!error.response || !error.response.data || !error.response.data.messages) {
```
bharathcs: Requires https://github.com/sivayogasubramanian/caza/pull/43#discussion_r970820485

```suggestion
function makeApiErrorResponse(error: AxiosError<ApiResponse<EmptyPayload>>): ApiResponse<EmptyPayload> {
  if (!error.response?.data?.messages) {
```

api/api.ts
```diff
@@ -1,8 +1,90 @@
...
+    .then((response: AxiosResponse) => {
+      const apiResponse = response.data;
+
+      if (process.env.NODE_ENV === 'development') {
+        console.info(`[API] ${response.status} ${endpoint} : ${getResponseMessages(apiResponse)}`);
+      }
+
+      return apiResponse;
+    })
+    .catch((error: AxiosError) => {
```
bharathcs: Needed for https://github.com/sivayogasubramanian/caza/pull/43#discussion_r970819855

```suggestion
    .catch((error) => {
```

pages/api/companies.ts
bharathcs: Resolved offline.

utils/url.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Converts an object into a URI query string.
+ * Supports arrays and nested objects.
```
bharathcs, line 3:
```suggestion
 * Converts an object into a URI query string.
 * Custom implementation required to ensure arrays and deeply nested objects are represented in a way handled by NextJS. 
 * e.g. the query params for array `foo=[1,2,3]` must be `?foo=1&foo=2&foo=3` not `?foo[]=1&foo[]=2&foo[]=3` as returned by Axios.
```

Btw I found a package online that handles this: https://stackoverflow.com/a/46153494, but I've briefly tested this and looks good! So if you want to stick with the implementation you wrote, I'm fine too.

bharathcs (approved):
LGTM! Just have some nits -- approving in advance here.

api/api.ts
sivayogasubramanian: This is just for the console log and will never be seen by the user.

api/api.ts
```diff
@@ -1,8 +1,90 @@
...
+        console.error(`[API] ${error.code} ${endpoint} : ${error.message}`);
+      }
+
+      throw makeApiErrorResponse(error);
+    });
+}
+
+function makeApiErrorResponse(error: AxiosError): ApiResponse<EmptyPayload> {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
ganhongyao: Can you briefly explain why we need to use `ts-ignore` for this and in line 80?

api/applicationStagesApi.ts
```diff
@@ -0,0 +1,34 @@
+import { ApiPromise, EmptyPayload } from '../types/apiResponse';
+import { ApplicationStageData, ApplicationStagePatchData, ApplicationStagePostData } from '../types/applicationStage';
+import api from './api';
+import { APPLICATIONS_API_ENDPOINT } from './applicationsApi';
+
+export const APPLICATION_STAGES_API_ENDPOINT = 'stages';
```
ganhongyao, line 6: Would it make more sense if we define a function `makeApplicationStagesApiEndpoint(applicationId: number)` instead? From what I see in the below functions we are never using this constant as it is, and we always have to do ``${APPLICATIONS_API_ENDPOINT}/${applicationId}/${APPLICATION_STAGES_API_ENDPOINT}``, so it would be good to extract that out.

api/tasksApi.ts
```diff
@@ -0,0 +1,24 @@
+import { ApiPromise, EmptyPayload } from '../types/apiResponse';
+import { TaskData, TaskPatchData, TaskPostData } from '../types/task';
+import api from './api';
+import { APPLICATIONS_API_ENDPOINT } from './applicationsApi';
+
+export const TASKS_API_ENDPOINT = 'tasks';
```
ganhongyao, line 6: Same comment as above

utils/url.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+// eslint-disable-next-line
+export function toQueryString(object: Record<string, any>, prefix = ''): string {
+  if (object === undefined || object === null) {
+    return '';
+  }
+
+  const query = Object.keys(object).map((key) => {
+    const value = object[key];
```
ganhongyao:
```suggestion
  const query = Object.entries(object).map(([key, value]) => {
```

ganhongyao:
image

The error seems to need some unpacking before it can be displayed. Is this out of scope of this PR?

sivayogasubramanian:
> image
> 
> The error seems to need some unpacking before it can be displayed. Is this out of scope of this PR?

I think we would unpack it in the toast service and display the error message to the user there. 

Edit: did it in this pr

api/api.ts
sivayogasubramanian: Typescript doesn't know that our data has a messages field. Do you have any better solutions?

api/api.ts
sivayogasubramanian: Let's not bother with this.

api/applicationStagesApi.ts
sivayogasubramanian: I think this is alright because we would need to differentiate `getApplication` and `getApplications` so its better to leave it as it is.

utils/url.ts
sivayogasubramanian: Let's just use toQueryString as I have tested it extensively.

api/api.ts
```diff
@@ -1,8 +1,90 @@
-import axios from 'axios';
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ganhongyao, line 1: Is this still needed?

api/api.ts
```diff
@@ -1,8 +1,90 @@
...
+  }
+
+  public delete<D>(url: string): ApiPromise<D> {
+    return processRequest(url, this.deleteData(url));
+  }
+}
+
+function processRequest<D>(endpoint: string, promise: AxiosPromise<ApiResponse<D>>): ApiPromise<D> {
+  return promise
+    .then((response: AxiosResponse) => {
```
ganhongyao:
```suggestion
    .then((response: AxiosResponse<ApiResponse<D>>) => {
```

Not required to compile, but I think it would still be good to add in case we need more type information about `apiResponse` in future. Currently `apiResponse` is inferred to be `any`, but with this change it is now inferred as `ApiResponse<D>`.

api/api.ts
ganhongyao:
```suggestion
    .catch((error: AxiosError<ApiResponse<D>>) => {
```

api/api.ts
```diff
@@ -1,8 +1,90 @@
...
+    .catch((error: AxiosError) => {
+      if (process.env.NODE_ENV === 'development') {
+        console.error(`[API] ${error.code} ${endpoint} : ${error.message}`);
+      }
+
+      throw makeApiErrorResponse(error);
+    });
+}
+
+function makeApiErrorResponse(error: AxiosError): ApiResponse<EmptyPayload> {
```
ganhongyao:
```suggestion
function makeApiErrorResponse<D>(error: AxiosError<ApiResponse<D>>): ApiResponse<D> {
```

ganhongyao: With all the above changes, we now have type information about `error.response`, which is of type `AxiosResponse<ApiResponse<D>>`. Then `error.response.data` will be of type `ApiResponse<D>` where the field `messages` is defined.

ganhongyao (changes requested):

api/api.ts
ganhongyao: See new thread

bharathcs: Doesn't this work?: https://github.com/sivayogasubramanian/caza/pull/43#discussion_r970819855

bharathcs: Nvm I realise we both suggested the same thing.

ganhongyao: Sorry, I missed out on your comment earlier.

Yup this works, but by principle the argument is actually not of that type. The argument that we pass is actually of `AxiosError<ApiResponse<D>>`, but the type parameter was lost because it was written as `.catch((error: AxiosError)` on line 55.

So doing https://github.com/sivayogasubramanian/caza/pull/43#discussion_r970819855 is actually hacky in some sense just to get around the `ts-ignore`

ganhongyao (approved):

PR #44, opened 2022-09-15
Implement header

sivayogasubramanian:
NOTE: I will make another PR to change styles.


components/header/Header.tsx
```diff
@@ -0,0 +1,84 @@
+import { GithubOutlined, LogoutOutlined } from '@ant-design/icons';
+import { Button } from 'antd';
+import { GithubAuthProvider, signInWithPopup, signOut } from 'firebase/auth';
+import { useContext } from 'react';
+import logo from '../../assets/placeholder.png';
```
bharathcs: minor nit: Could we rename the file to `logoPlaceholder.png`?

components/header/Header.tsx
```diff
@@ -0,0 +1,84 @@
...
+      });
+      return;
+    }
+
+    const provider = new GithubAuthProvider();
+    provider.setCustomParameters({
+      allow_signup: 'false',
+    });
+
+    signInWithPopup(auth, provider).then((userCredentials) =>
```
bharathcs: `signInWithRedirect` is preferred over pop up for interactions with mobile browsers / PWA.

I can't find the source right now, so I can't say why its the best practice but I think its because some browsers are pop up hostile (like safari) and mobile experience doesn't handle pop ups consistently / smoothly.

components/header/Header.tsx
```diff
@@ -0,0 +1,84 @@
...
+      openNotification({ type: StatusMessageType.SUCCESS, message: `Welcome, ${userCredentials.user.displayName}!` }),
```
bharathcs: does `user.displayName` work for you? It just returns null for me.

bharathcs: Follow up: my GitHub provider does not supply name:
image

image

components/header/Header.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <div className="pl-5 pr-5 md:pl-20 md:pr-20 pt-3 pb-3 flex justify-between items-center sticky top-0 bg-slate-100">
+      <img src={logo.src} width="175px" />
+
+      {currentUser?.isAnonymous && (
+        <Button
+          type="primary"
+          icon={<GithubOutlined />}
+          className="items-center flex bg-blue-400 text-black rounded-md hover:bg-blue-500 hover:text-black focus:text-black"
```
bharathcs: Not sure if you count this as part of the next PR to fix styles but just in case it was missed:

Do we intend there to be a border of a different colour around the button in both states?

at rest:
image

on hover:
image

bharathcs: Ok turns out the border is from antd blue (rgb(96, 165, 250)), while the inside is from tailwind blue-4/500 (rgb(64, 169, 255) / rgb(24, 144, 255))

I'm not rly sure what's the best way to force antd / tailwind to match globally, any suggestions?

components/header/Header.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <Button
+            type="primary"
+            icon={<LogoutOutlined />}
+            className="items-center flex bg-blue-400 text-black rounded-md hover:bg-blue-500 hover:text-black focus:text-black"
+            onClick={handleLogout}
+          >
+            Log out
+          </Button>
+
+          <img src={githubProviderData?.photoURL || profilePlaceholder.src} width="35px" className="rounded-full" />
```
bharathcs: I personally prefer not showing a photo if GitHub doesn't provide a url. Wdyt?

bharathcs (commented):
Nice, functionality works for me!

components/header/Header.tsx
sivayogasubramanian: Your github does not have a display name lol. I think we have to handle this somehow.

components/header/Header.tsx
sivayogasubramanian: I'll remove the border for both.

components/header/Header.tsx
sivayogasubramanian: We cannot show the user message if we use `signInWithRedirect`. So ill remove it for this method.

components/header/Header.tsx
```diff
@@ -43,7 +40,7 @@ function Header() {
     signOut(auth).then(() =>
       openNotification({
         type: StatusMessageType.SUCCESS,
-        message: `See you again soon, ${currentUser?.displayName}!`,
+        message: currentUser?.displayName ? `See you again soon, ${currentUser?.displayName}!` : 'See you again soon!',
```
bharathcs: minor nit
```suggestion
        message: currentUser?.displayName ? `See you again soon, ${currentUser.displayName}!` : 'See you again soon!',
```

components/header/Header.tsx
```diff
@@ -56,7 +53,7 @@ function Header() {
         <Button
           type="primary"
           icon={<GithubOutlined />}
-          className="items-center flex bg-blue-400 text-black rounded-md hover:bg-blue-500 hover:text-black focus:text-black"
+          className="items-center flex bg-blue-400 text-black rounded-md hover:bg-blue-500 hover:text-black focus:text-black border-none hover:border-none"
```
bharathcs, line 56: damn here I was with my smol brain. 

Good fix. 👍

bharathcs (approved):
LGTM!

PR #45, opened 2022-09-15
Add roles/world GET handler

ganhongyao:
Changes
- Add GET handler for `roles/world` endpoint
- Extract out `convertQueryParamToStringArray` method to reduce code duplication

https://user-images.githubusercontent.com/75519136/190895995-541ce13e-9718-49bf-971f-0b9ebcb5e6f6.mp4

### DIFF
--- a/pages/api/roles/index.ts
+++ b/pages/api/roles/index.ts
@@ -5,7 +5,13 @@ import { RoleData, RoleListData, RolePostData, RoleQueryParams } from '../../../
 import { Nullable } from '../../../types/utils';
 import { withAuth } from '../../../utils/auth/jwtHelpers';
 import { MIN_YEAR } from '../../../utils/constants';
-import { createJsonResponse, HttpMethod, HttpStatus, rejectHttpMethod } from '../../../utils/http/httpHelpers';
+import {
+  convertQueryParamToStringArray,
+  createJsonResponse,
+  HttpMethod,
+  HttpStatus,
+  rejectHttpMethod,
+} from '../../../utils/http/httpHelpers';
 import { withPrismaErrorHandling } from '../../../utils/prisma/prismaHelpers';
 import { capitalizeEveryWord, splitByWhitespaces } from '../../../utils/strings/formatters';
 import { isEmpty } from '../../../utils/strings/validations';
@@ -122,12 +128,10 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<
 }
 
 function parseGetQueryParams(req: NextApiRequest): RoleQueryParams {
-  const companyId = req.query.companyId;
-  const searchQuery = req.query.searchWords;
-  const searchWords =
-    searchQuery === undefined ? [] : Array.isArray(searchQuery) ? searchQuery : splitByWhitespaces(searchQuery);
+  const { companyId, searchWords } = req.query;
+  const searchWordsArr = convertQueryParamToStringArray(searchWords, splitByWhitespaces);
   const parsedCompanyId = canBecomeInteger(companyId) ? Number(companyId) : undefined;
-  return { companyId: parsedCompanyId, searchWords };
+  return { companyId: parsedCompanyId, searchWords: searchWordsArr };
 }
 
 function validatePostRequest(req: NextApiRequest): Nullable<MessageType> {
--- a/pages/api/roles/world/index.ts
+++ b/pages/api/roles/world/index.ts
@@ -3,7 +3,7 @@ import type { NextApiRequest, NextApiResponse } from 'next';
 import { ApiResponse } from '../../../../types/apiResponse';
 import { WorldRoleListData, WorldRoleQueryParams } from '../../../../types/role';
 import { getNonEmptyArrayOrUndefined, combineDefinedArrays, buildFrequencyMap } from '../../../../utils/arrays';
-import { withVerifiedUser } from '../../../../utils/auth/jwtHelpers';
+import { withVerified } from '../../../../utils/auth/jwtHelpers';
 import {
   makeCompanyNameFilters,
   makeRoleTitleFilters,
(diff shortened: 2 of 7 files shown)


pages/api/roles/index.ts
```diff
@@ -122,12 +128,10 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse<ApiResponse<
 }
 
 function parseGetQueryParams(req: NextApiRequest): RoleQueryParams {
-  const companyId = req.query.companyId;
-  const searchQuery = req.query.searchWords;
-  const searchWords =
-    searchQuery === undefined ? [] : Array.isArray(searchQuery) ? searchQuery : splitByWhitespaces(searchQuery);
+  const { companyId, searchWords } = req.query;
+  const searchWordsArr = convertQueryParamToStringArray(searchWords, splitByWhitespaces);
```
zognin, line 132: Could you help me do a similar change for applications? Thank you

pages/api/roles/world/index.ts
```diff
@@ -1,5 +1,105 @@
...
+      type: true,
+      year: true,
+      company: {
+        select: {
+          id: true,
+          name: true,
+          companyUrl: true,
+        },
+      },
+      applications: {
```
zognin, line 64: Do we need to check that the applications belong to verified (non-anonymous) users only?

pages/api/roles/world/index.ts
```diff
@@ -1,5 +1,105 @@
...
+  const searchWordsArr = convertQueryParamToStringArray(searchWords, splitByWhitespaces);
+  const roleTypeUncheckedWords = convertQueryParamToStringArray(roleTypes, splitByCommaRemovingWhitespacesAround);
+
+  // Safe to typecast due to the filter check.
+  const roleTypeWords: RoleType[] = roleTypeUncheckedWords.filter((word) => word in RoleType) as RoleType[];
+
+  return { searchWords: searchWordsArr, roleTypeWords };
+}
+
+export default withPrismaErrorHandling(withAuthUser(handler));
```
zognin: I think only non anonymous users can fetch these right?
```suggestion
export default withPrismaErrorHandling(withVerifiedUser(handler));
```

zognin (commented):
Looks good, have some questions.

pages/api/roles/world/index.ts
ganhongyao: Currently with our database schema there is no way to tell whether a user is anonymous, although we could support it rather easily.

But either way, I think we don't need to check that since if we filter maybe the data would be lacking. Is your concern about spammy data?

utils/arrays.ts
```diff
@@ -5,3 +5,11 @@ export function combineDefinedArrays<T>(items: (T[] | undefined)[]) {
 export function getNonEmptyArrayOrUndefined<T>(items: T[]) {
   return items.length === 0 ? undefined : items;
 }
+
+export function buildFrequencyMap<T>(items: T[]): Map<T, number> {
```
bharathcs, line 14: Add warning.

```suggestion
/**
 * Builds frequency map of items. Note that Javascript Map is only guaranteed to compare keys by its value if it is a
 * primitive type or a wrapper for a primitive type. For non-primitive types used as keys, Map will use object value
 * comparison ( { foo: true } !== { foo: true } )
 */
export function buildFrequencyMap<T>(items: T[]): Map<T, number> {
```

bharathcs: If you feel like requiring `T extends (string | number | boolean)` instead of the warning I'm fine with that too.

pages/api/roles/world/index.ts
```diff
@@ -1,5 +1,105 @@
...
+  const method = req.method;
+  switch (method) {
+    case HttpMethod.GET:
+      return handleGet(userId, req, res);
+    default:
+      return rejectHttpMethod(res, method);
+  }
+}
+
+async function handleGet(userId: string, req: NextApiRequest, res: NextApiResponse<ApiResponse<WorldRoleListData[]>>) {
```
bharathcs, line 35:
```suggestion
async function handleGet(req: NextApiRequest, res: NextApiResponse<ApiResponse<WorldRoleListData[]>>) {
```

You'll need to change handler method signature as well.

pages/api/roles/world/index.ts
```diff
@@ -1,5 +1,105 @@
...
+import { withPrismaErrorHandling } from '../../../../utils/prisma/prismaHelpers';
+import { splitByWhitespaces, splitByCommaRemovingWhitespacesAround } from '../../../../utils/strings/formatters';
 
-export default function handler(req: NextApiRequest, res: NextApiResponse) {
-  res.status(200).end(`Request sent to ${req.url}\n`);
+const prisma = new PrismaClient();
+
+async function handler(userId: string, req: NextApiRequest, res: NextApiResponse) {
+  const method = req.method;
+  switch (method) {
```
bharathcs, line 27: Super nit (sorry)
```suggestion
  switch (req.method) {
```

bharathcs (approved):
LGTM, thanks for cleaning up the logic! I suggest we merge this first, and handle https://github.com/sivayogasubramanian/caza/pull/45#discussion_r972544740 (using anonymous data discussion) later, because it would require a db change.

pages/api/roles/world/index.ts
ganhongyao: I think this is fine, since we are also using `method` to pass to `rejectHttpMethod`, rather than doing `req.method` at both places.

pages/api/roles/world/index.ts
ganhongyao: Can you help me verify that this https://github.com/sivayogasubramanian/caza/pull/45/commits/ee9528844a5519e421164ef8285fef11edf2c46d is what you mean by changing the handler method signature? We will still need the first string parameter (albeit unused) since we are using `withVerifiedUser` here.

bharathcs: Actually the current main has a new middleware `withVerified`, could you merge and use that?

bharathcs (approved):
LGTM, suggest merging this and discussing the unverified data question in a later PR.

sivayogasubramanian (approved):

PR #46, opened 2022-09-15
Add applications list search bar and filters

ganhongyao:
Changes
- TextField
- StageType Dropdown
- RoleType Dropdown

https://user-images.githubusercontent.com/75519136/190494913-70f92d54-0197-43da-9e6c-6b141fd9cef3.mp4


types/role.ts
```diff
@@ -1,6 +1,14 @@
 import { RoleType } from '@prisma/client';
 import { CompanyData } from './company';
 
+export const RoleTypeToLabelMap = Object.freeze({
+  [RoleType.WINTER_INTERNSHIP]: 'Winter Internship',
```
sivayogasubramanian: Let's use immutable map here.

types/applicationStage.ts
```diff
@@ -1,6 +1,18 @@
 import { ApplicationStageType } from '@prisma/client';
 import { Nullable } from './utils';
 
+export const ApplicationStageTypeToLabelMap = Object.freeze({
+  [ApplicationStageType.APPLIED]: 'Applied',
```
sivayogasubramanian: Same

pages/applications/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  const onApplicationStageTypesFilterChange = (stageTypes: ApplicationStageType[]) => {
+    setSearchParams({ ...searchParams, stageTypeWords: stageTypes });
+  };
+
+  const { data } = useSWR<ApiResponse<ApplicationListData[]>>(
+    [APPLICATIONS_API_ENDPOINT, searchParams],
+    (url, searchParams) => api.get(url, { params: searchParams }),
+  );
```
sivayogasubramanian: Are we able to debounce this or throttle this? I think throttle would be better here.

PR #47, opened 2022-09-16
Require payload member types to be string or number.

bharathcs:
- Set type standard for payloads.
- Update `BaseApi`.
- Update api/app/[:appId]/tasks
- Update stages type.
- Update api/app/:appId/stages
- Update api/app/:appId/stages/:stageId.
- Refactor to have a single converter.
- Create TaskData converter.
- Update api/applications.
- Put in explicit types for handlers.

### DIFF
--- a/pages/api/applications/index.ts
+++ b/pages/api/applications/index.ts
@@ -5,7 +5,7 @@ import {
   ApplicationQueryParams,
   ApplicationRoleData,
 } from '../../../types/application';
-import { ApplicationStageType, Prisma, PrismaClient, RoleType } from '@prisma/client';
+import { ApplicationStage, ApplicationStageType, Company, Prisma, PrismaClient, Role, RoleType } from '@prisma/client';
 import { createJsonResponse, HttpMethod, HttpStatus, rejectHttpMethod } from '../../../utils/http/httpHelpers';
 import { withAuthUser } from '../../../utils/auth/jwtHelpers';
 import { ApiResponse, StatusMessageType } from '../../../types/apiResponse';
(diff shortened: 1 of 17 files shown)


utils/task/converter.ts
```diff
@@ -0,0 +1,13 @@
+import { Task } from '@prisma/client';
+import { TaskData } from '../../types/task';
+
+export function convertTaskToPayload(task: Task): TaskData {
+  const { id, title, dueDate, notificationDateTime, isDone } = task;
+  return {
+    id,
+    title,
+    dueDate: dueDate.toJSON() ?? null,
```
zognin: I don't think dueDate is supposed to be null?
```suggestion
    dueDate: dueDate.toJSON(),
```

types/applicationStage.ts
```diff
@@ -28,6 +28,6 @@ export type ApplicationStagePatchData = {
 export type ApplicationStageApplicationListData = {
   id: number;
   type: ApplicationStageType;
-  date: Date;
+  date: string;
   emojiUnicodeHex: Nullable<string>;
 };
```
zognin, line 33: I think we wanted to change post and patch data date types to string too? `ApplicationStagePatchData` and `ApplicationStagePostData` weren't changed yet.

pages/api/applications/index.ts
```diff
@@ -240,4 +216,30 @@ function validatePostRequest(req: NextApiRequest): Nullable<MessageType> {
...
+    id,
+    taskNotificationCount,
+    role: {
+      type,
+      title,
+      year,
+      id: role.id,
+      company: { id: company.id, name: company.name, companyUrl: company.companyUrl },
+    },
+    latestStage: convertApplicationStageToPayload(latestStage),
```
zognin: `latestStage` may be undefined if an application has no stages.
Screenshot 2022-09-16 at 12 29 35 PM

zognin (commented):
Mostly looks good thanks for doing this

utils/task/converter.ts
bharathcs: Oh, thanks for the catch!

zognin (approved):
LGTM thanks!

PR #55, opened 2022-09-17
Add application create forms

ganhongyao:

PR #56, opened 2022-09-17
Add task modal

zognin:
Stuff added
- Generic Task form
- Add task modal
- Edit task modal
- Temporary add task button

Note
- Please ignore the add task button, it's just there to test the add new task form
- Please ignore the white primary buttons. Not changing theme in this PR.
- **BUG**: There is a timezone related bug when using moment but I am too tired to figure it out now. Will fix later.

Questions
- Should `EditTaskModal` automatically close after a successful save?


components/cards/ApplicationTaskTimelineCard.tsx
```diff
@@ -5,15 +5,16 @@ import { Checkbox } from 'antd';
...
 import { CheckboxChangeEvent } from 'antd/lib/checkbox';
 import tasksApi from '../../api/tasksApi';
-import { Dispatch, SetStateAction } from 'react';
+import { Dispatch, MouseEventHandler, SetStateAction } from 'react';
 
 interface Props {
   applicationId: number;
   task: TaskData;
   setShouldFetchData: Dispatch<SetStateAction<boolean>>;
+  onClick: MouseEventHandler<HTMLDivElement> | undefined;
```
sivayogasubramanian:
```suggestion
  onClick?: () => void;
```

components/modals/EditTaskModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+const { confirm } = Modal;
+
+interface Props {
+  applicationId: number;
+  initialTask: TaskData;
+  setSelectedTask: React.Dispatch<React.SetStateAction<Nullable<TaskData>>>;
+  setShouldFetchData: Dispatch<SetStateAction<boolean>>;
+}
+
+function EditTaskModal({ applicationId, initialTask, setSelectedTask, setShouldFetchData }: Props) {
```
sivayogasubramanian: The close button does not work.

https://user-images.githubusercontent.com/53614454/190885852-99bf28ad-a897-4501-bd6e-b3f4c766a47a.mp4

utils/task/taskUtils.ts
```diff
@@ -0,0 +1,26 @@
+import { NotificationDateTimeType, TaskFormData } from '../../types/task';
+import moment from 'moment';
+
+export const DEFAULT_NOTIFICATION_TIME = moment('09:00:00', 'hh:mm:ss');
+export const DEFAULT_NOTIFICATION_DAYS_OFFSET = 1;
+
+export function getNotificationDateTime(values: TaskFormData) {
```
sivayogasubramanian: Is it possible to return a moment object here? The logic of this function is a bit hard to understand and it couples our times to whatever date string format this is returning.

https://stackoverflow.com/questions/42404507/moment-js-concatenate-date-and-time

utils/date/formatters.ts
```diff
@@ -24,6 +23,11 @@ export function getCountOfDaysTillTodayFrom(date: Date): number {
   return Math.ceil(differenceInTime / (1000 * 3600 * 24));
 }
 
+export function calculateDaysOffset(currentDate: Date, referenceDate: Date) {
```
sivayogasubramanian, line 25: Could you help to use this for the function in line 22? Thanks :)

components/modals/NewTaskModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+  applicationId: number;
+  setIsAddingNewTask: React.Dispatch<React.SetStateAction<boolean>>;
+  setShouldFetchData: Dispatch<SetStateAction<boolean>>;
+}
+
+function NewTaskModal({ applicationId, setIsAddingNewTask, setShouldFetchData }: Props) {
+  const [shouldDisableSaveButton, setShouldDisableSaveButton] = useState(true);
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [taskFormData, setTaskFormData] = useState<Nullable<TaskFormData>>(null);
+  const [initialValues] = useState<TaskFormData>({
```
sivayogasubramanian, line 25: Why does this use useState when the state is never changed?

components/forms/TaskForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+import React, { useEffect, useState } from 'react';
+import { NotificationDateTimeType, TaskFormData } from '../../types/task';
+import { Nullable } from '../../types/utils';
+
+interface Props {
+  initialValues: TaskFormData;
+  shouldTouchAllCompulsoryFields: boolean;
+  shouldShowMarkDone: boolean;
+  isSubmitting: boolean;
+  setShouldDisableSaveButton: React.Dispatch<React.SetStateAction<boolean>>;
```
sivayogasubramanian: I think we should not disable the button. Right now the user has no way of knowing why the button is disabled. Although I am able to infer that my compulsory fields are not filled, I think it's just better to let the validations take over. 

Right now the validations in this form never get shown.

Also just to implement this feature, we are doing a lot of things which makes this way more complicated than it should be.

components/forms/TaskForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const [form] = Form.useForm();
+
+  const [shouldShowNotificationDaysOffsetInput, setShouldShowNotificationDaysOffsetInput] = useState(false);
+  const [shouldShowNotificationDateTimePicker, setShouldShowNotificationDateTimePicker] = useState(false);
+  const [shouldShowNotificationTimePicker, setShouldShowNotificationTimePicker] = useState(false);
+
+  useEffect(() => {
+    form.resetFields();
+    onSelectNotificationDateTimeType(initialValues.notificationDateTimeType ?? NotificationDateTimeType.NONE);
+  }, [initialValues]);
```
sivayogasubramanian, line 28:
```suggestion
  }, []);
```

Am I right to say that `initialValues` will never be changed after this component mounts?

components/forms/TaskForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+          )}
+
+          <Form.Item name="notificationDateTimeType" className="flex-grow">
+            <Select onSelect={onSelectNotificationDateTimeType}>
+              <Option value={NotificationDateTimeType.NONE}>None</Option>
+              <Option value={NotificationDateTimeType.DAY_OF_EVENT}>On day of event</Option>
+              <Option value={NotificationDateTimeType.DAYS_BEFORE}>Day(s) before</Option>
+              <Option value={NotificationDateTimeType.DAYS_AFTER}>Day(s) after</Option>
+              <Option value={NotificationDateTimeType.ON_SELECTED_DATE}>On selected date</Option>
+            </Select>
```
sivayogasubramanian: Should we extract this out?

Refer to `RoleTypesSelect`.

components/forms/TaskForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const getTimePickerComponent = (givenStyle?: React.CSSProperties) => (
+    <Form.Item
+      name="notificationTime"
+      style={givenStyle && givenStyle}
+      rules={[{ required: true, message: 'Please select a time.' }]}
+    >
+      <TimePicker style={{ width: '100%' }} />
+    </Form.Item>
+  );
```
sivayogasubramanian: Let's make this a generic TimePicker component.

components/modals/EditTaskModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+import tasksApi from '../../api/tasksApi';
+import {
+  DEFAULT_NOTIFICATION_DAYS_OFFSET,
+  DEFAULT_NOTIFICATION_TIME,
+  getNotificationDateTime,
+} from '../../utils/task/taskUtils';
+import { Button, Modal } from 'antd';
+import { ExclamationCircleOutlined } from '@ant-design/icons';
+
+const { confirm } = Modal;
```
sivayogasubramanian: You can just do `Modal.confirm`. All these destructing makes it seem like confirm is a function that you created.

components/modals/EditTaskModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+interface Props {
+  applicationId: number;
+  initialTask: TaskData;
+  setSelectedTask: React.Dispatch<React.SetStateAction<Nullable<TaskData>>>;
+  setShouldFetchData: Dispatch<SetStateAction<boolean>>;
```
sivayogasubramanian: Can standardize the notation?

I prefer the latter.

components/modals/EditTaskModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+  };
+
+  const onDelete = () => {
+    confirm({
+      title: 'Are you sure about deleting this task?',
+      icon: <ExclamationCircleOutlined />,
+      content: 'This action is irreversible',
+      onOk() {
+        handleDelete();
+      },
```
sivayogasubramanian:
```suggestion
      onOk: handleDelete,
```

components/modals/EditTaskModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+      notificationDateTime: getNotificationDateTime(values),
+      isDone: values.isDone,
+    };
+
+    tasksApi
+      .editTask(applicationId, task.id, taskPatchData)
+      .then((value) => {
+        const updatedTask = value.payload as TaskData;
+        setTask(updatedTask);
+        setShouldFetchData(true);
```
sivayogasubramanian: Instead of doing this, is it better to just mutate instead?

https://swr.vercel.app/docs/mutation

Refer to `pages/applications/create.tsx` of HY PR line 92.

components/modals/NewTaskModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const taskPostData: TaskPostData = {
+      title: values.title ?? 'Task',
+      dueDate: values.dueDate?.toISOString() ?? new Date().toISOString(),
+      notificationDateTime: notificationDateTime === undefined ? new Date().toISOString() : notificationDateTime,
+    };
+
+    tasksApi
+      .createTask(applicationId, taskPostData)
+      .then(() => {
+        setShouldFetchData(true);
```
sivayogasubramanian: Same here.

components/modals/NewTaskModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+  DEFAULT_NOTIFICATION_TIME,
+  getNotificationDateTime,
+} from '../../utils/task/taskUtils';
+import { Nullable } from '../../types/utils';
+import { Modal } from 'antd';
+
+interface Props {
+  applicationId: number;
+  setIsAddingNewTask: React.Dispatch<React.SetStateAction<boolean>>;
+  setShouldFetchData: Dispatch<SetStateAction<boolean>>;
```
sivayogasubramanian: Same here.

components/modals/EditTaskModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+  setSelectedTask: React.Dispatch<React.SetStateAction<Nullable<TaskData>>>;
+  setShouldFetchData: Dispatch<SetStateAction<boolean>>;
+}
+
+function EditTaskModal({ applicationId, initialTask, setSelectedTask, setShouldFetchData }: Props) {
+  const [task, setTask] = useState(initialTask);
+  const [initialValues, setInitialValues] = useState<TaskFormData>({
+    notificationDaysOffset: DEFAULT_NOTIFICATION_DAYS_OFFSET,
+  });
+  const [shouldDisableSaveButton, setShouldDisableSaveButton] = useState(true);
```
sivayogasubramanian: Let's standarize to use generics for inferable types.

But this is up to you to decide.

This applies to all other files as well.

components/modals/NewTaskModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }, [taskFormData]);
+
+  const onCancel = () => {
+    setIsAddingNewTask(false);
+  };
+
+  const onSubmit = () => {
+    setIsSubmitting(true);
+  };
```
sivayogasubramanian: All these functions adds extra clutter to this file. Can we just do `() => setIsAddingNewTask(false)`

components/modals/NewTaskModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const submit = (values: TaskFormData) => {
+    const notificationDateTime = getNotificationDateTime(values);
+
+    // Note: notificationDateTime can be null
+    const taskPostData: TaskPostData = {
+      title: values.title ?? 'Task',
+      dueDate: values.dueDate?.toISOString() ?? new Date().toISOString(),
+      notificationDateTime: notificationDateTime === undefined ? new Date().toISOString() : notificationDateTime,
```
sivayogasubramanian:
```suggestion
      notificationDateTime: notificationDateTime === undefined ? null : notificationDateTime,
```

Should this be null instead according to you comment?

pages/applications/[applicationId].tsx
```diff
@@ -71,14 +76,37 @@ function Application() {
...
+      )}
+
+      {selectedTask && applicationId && (
+        <EditTaskModal
+          applicationId={applicationId}
+          initialTask={selectedTask}
+          setSelectedTask={setSelectedTask}
+          setShouldFetchData={setShouldFetchData}
+        />
+      )}
```
Comment (sivayogasubramanian) [inline pages/applications/[applicationId].tsx]: Is it better for the `XModal` to take in a `isOpen` prop and we use a useState here for the state management.

We can also return null if `applicationId` is not defined.

Suggesting based on HY's approach in `pages/applications/create.tsx` in https://github.com/sivayogasubramanian/caza/pull/55

components/modals/NewTaskModal.tsx
zognin: useState isn't used only when the state will change. It also ensures it doesn't keep creating a new object unless necessary. You can try removing it and there will be a bug.

components/forms/TaskForm.tsx
zognin: No.

components/modals/EditTaskModal.tsx
zognin: This is just telling the parent component to mutate, the parent component is still using mutate. Unless you want me to pass the mutate function down to the child? Ref to `/[applicationId].tsx` line 72.

If you're talking about mutating task, Task does not have a GET endpoint.

components/modals/NewTaskModal.tsx
zognin: It can be anything because the undefined case shouldn't happen (by right should throw error), I can change to null if that's clearer.

components/forms/TaskForm.tsx
zognin: Not going to do it at the moment. I think it should be extracted only if it improves clarity or it will be reused elsewhere. Currently extracting these reduces clarity.

pages/applications/[applicationId].tsx
Comment (zognin) [reply pages/applications/[applicationId].tsx]: Rather not because once that happens I can pass in nullable task and application id and that makes the checks in `EditTaskModal` more messy.

components/modals/NewTaskModal.tsx
zognin: I kept one of these (in `EditTaskModal`) because it's used in 2 areas

utils/date/formatters.ts
```diff
@@ -20,7 +19,11 @@ export function makeDisplayNotificationDatetime(notificationDatetime: Date, dueD
 }
 
 export function getCountOfDaysTillTodayFrom(date: Date): number {
-  const differenceInTime = Date.now() - date.getTime();
+  return calculateDaysOffset(new Date(), date);
```
sivayogasubramanian: The logic is not the same as before.

components/forms/TaskForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+  isSubmitting: boolean;
+  setTaskFormData: Dispatch<SetStateAction<Nullable<TaskFormData>>>;
+}
+
+interface TaskFormTimePickerProps {
+  isOpen: boolean;
+  givenStyle?: CSSProperties;
+}
+
+const { Option } = Select;
```
sivayogasubramanian: Let's use Select.Option here

components/forms/TaskForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+  useEffect(() => {
+    form.resetFields();
+    onSelectNotificationDateTimeType(initialValues.notificationDateTimeType ?? NotificationDateTimeType.NONE);
+  }, [initialValues]);
+
+  useEffect(() => {
+    if (isSubmitting) {
+      setTaskFormData(form.getFieldsValue());
+    }
+  }, [isSubmitting]);
```
sivayogasubramanian: Validation does not work.

```suggestion
  useEffect(() => {
    if (isSubmitting) {
      form.validateFields().then(() => setTaskFormData(form.getFieldsValue()));
    }
  }, [isSubmitting]);
```

sivayogasubramanian (approved):
LGTM! Pre-approving. You can merge when the issues bellow are resolved.

PR #64, opened 2022-09-18
Fix contributors logic

ganhongyao:
## Context

Currently, there is a `contributorId` field in the `Role` and `Company` models. @zognin pointed out that this does not work because for a company that is still not verified, users that try to "create" this same company will not be able to see the company (because they are not the first contributor) but yet they will still see the message "Company already exists".

## Changes
- Make `Role <-> Contributor` and `Company <> Contributor` a many-to-many relationship.
- **GET** logic
   - Return all companies that are verified or I am one of the contributors
- **CREATE** logic
   - Check if company exists.
   - If company does not exist, create and return company.
   - If the company is already verified or I am already a contributor, return company.
   - Else, add me as a contributor and return company.


prisma/migrations/20220918154615_make_contributions_many_to_many/migration.sql
```diff
@@ -0,0 +1,82 @@
...
+-- AddForeignKey
+ALTER TABLE "companies_contributions" ADD CONSTRAINT "companies_contributions_contributor_id_fkey" FOREIGN KEY ("contributor_id") REFERENCES "users"("uid") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "roles_contributions" ADD CONSTRAINT "roles_contributions_role_id_fkey" FOREIGN KEY ("role_id") REFERENCES "roles"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "roles_contributions" ADD CONSTRAINT "roles_contributions_contributor_id_fkey" FOREIGN KEY ("contributor_id") REFERENCES "users"("uid") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- Trigger to verify a company once there are >= 5 contributors
```
ganhongyao: Code below here is authored by me, might want to manually review this.

pages/api/companies.ts
```diff
@@ -95,31 +102,77 @@ async function handlePost(userId: string, req: NextApiRequest, res: NextApiRespo
...
+    return;
+  }
 
-  if (duplicateCompany) {
+  const isAlreadyContributedByUser = duplicateCompany.contributions.some(
+    (contribution) => contribution.contributorId === userId,
+  );
+
+  if (duplicateCompany.isVerified || isAlreadyContributedByUser) {
     res.status(HttpStatus.OK).json(createJsonResponse(duplicateCompany, messages[MessageType.COMPANY_ALREADY_EXISTS]));
```
zognin: do we need to update the type `CompanyData` if we pass in the entire `duplicateCompany`? But `createdCompany` doesn't have the extra fields. I think we can take only the fields needed from `duplicateCompany`.

prisma/migrations/20220918154615_make_contributions_many_to_many/migration.sql
```diff
@@ -0,0 +1,82 @@
...
+ALTER TABLE "roles" DROP CONSTRAINT "roles_contributor_id_fkey";
+
+-- AlterTable
+ALTER TABLE "companies" DROP COLUMN "contributor_id";
+
+-- AlterTable
+ALTER TABLE "roles" DROP COLUMN "contributor_id";
+
+-- CreateTable
+CREATE TABLE "companies_contributions" (
```
zognin: Is the table supposed to be 
```suggestion
CREATE TABLE "company_contributions" (
```

Same for `role_contributions`?

zognin (approved):
LGTM, some nits, thanks!

ganhongyao:
@sivayogasubramanian can you check that the reason for the deployment failure is because of the migrations? I deleted and manually edited the old migration files in https://github.com/sivayogasubramanian/caza/pull/64/commits/77a62d486d7c16c7490392de3c6e486217c9f012 as I didn't want the old trigger to lie around.

If that's the reason then help me force merge this thanks :)

PR #65, opened 2022-09-18
Add stage modal

### LINKED ISSUE #41: Validate stage order in API calls once validation functions are confirmed.
Blocked by #36 

Checks required at:

- [x] POST stages
- [x] PATCH stages/:stageID
- [x] DELETE stages/:stageId

After this, only direct prisma / Postgres action should be able to cause stage orders to go bad.

zognin:
Things added
- New stage modal
- Edit stage modal
- General stage form
- Temporary add stage button
- Fixed some task form bugs too

Note
- Run `yarn install` to install new dependency: [emoji-picker-react](https://www.npmjs.com/package/emoji-picker-react). It's a bit slow sometimes. And it requires dynamic import because it's very big.
- Some emojis render funny.
photo_2022-09-19 01 48 20


components/cards/ApplicationStageTimelineCard.tsx
```diff
@@ -5,13 +5,14 @@ import { isValidDate } from '../../utils/date/validations';
...
+  onClick?: () => void;
 }
 
-function ApplicationStageTimelineCard({ applicationStage }: Props) {
+function ApplicationStageTimelineCard({ applicationStage, onClick }: Props) {
   const date = isValidDate(applicationStage.date) ? new Date(applicationStage.date) : undefined;
 
   return (
-    <div className="shadow-md rounded-lg">
+    <div className="shadow-md rounded-lg" onClick={onClick}>
```
ganhongyao: Should we make the cursor a pointer and perhaps darken the background when the user is hovering over the div? To make it clear that it is clickable (for Web users).

components/forms/StageForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (isSubmitting) {
+      form
+        .validateFields()
+        .then(() => setStageFormData({ ...form.getFieldsValue(), emojiUnicodeHex: selectedEmojiUnicode }))
+        .catch(() => setIsSubmitting(false));
+    }
+  }, [isSubmitting]);
+
+  return (
+    <Form form={form} initialValues={initialValues}>
```
ganhongyao: image

UI nit but can we align the start of the fields?
```suggestion
    <Form form={form} initialValues={initialValues} labelCol={{ span: 8 }} wrapperCol={{ span: 16}}>
```

Something like that should work, though the `span` values should be adjusted.

components/forms/StageForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <Form.Item name="type" label="Stage" rules={[{ required: true, message: 'Please choose a stage.' }]}>
+        <Select placeholder="Select stage">
+          {stageTypeToDisplayStringMap.map((value, key) => (
```
ganhongyao, line 53: The order of the stages right now is random, because of the `Immutable.Map`, which can make it quite confusing for the user.

Should we use a [Immutable.OrderedMap](https://immutable-js.com/docs/v4.1.0/OrderedMap/) for `stageTypeToDisplayStringMap` instead?

components/modals/EditStageModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+    });
+  }, [stage]);
+
+  const onCancel = () => setSelectedStage(null);
+
+  const onDelete = () => {
+    Modal.confirm({
+      title: 'Are you sure about deleting this stage?',
+      icon: <ExclamationCircleOutlined />,
+      content: 'This action is irreversible',
```
ganhongyao:
```suggestion
      content: 'This action is irreversible!',
```
Nit

ganhongyao (commented):
I also realised that the `validateStageChronology` is not integrated into our app logic and the chronology can be messed up.

Should we do that in this PR?

ganhongyao (approved):
Pre-approving, feel free to merge once addressed

PR #72, opened 2022-09-19
Add buttons on application page

zognin:
- The "role stats" button will only show if the role is verified. If the user is not verified, then the detailed world page needs to do the blurring.


bharathcs:
Minor thing: I realise our text colours are inconsistent across blue buttons:

image

bharathcs (approved):
The colours comment I made can be handled when you're working on the theme stuff.

sivayogasubramanian (approved):

PR #79, opened 2022-09-20
Create user on anonymous log in, and link if possible on github log in.

bharathcs:
# Test Plan

## Basic Test (plz halp)

1. Open prisma studio
1. Clear ur cache / open in incognito 
1. Verify new anonymous user has been added.
1. Create application should succeed.
1. Verify on prisma studio that the new application is added
1. Log in to Github
1. Reload page (required because of bug #53)
1. Verify the application you made still exists.
1. Verify on prisma studio that the application has been transferred to the new GitHub uid, and old UID is missing.

## A bit more testing (optional, if you would be so kind):

1. Log out from GitHub (Now should be anonymous)
1. Verify on prisma studio that there is a new UID.
1. Reload page (required because of bug #53)
1. Verify no applications present in page
1. Log in to Github
1. This should succeed, but there should be an alert saying "There is already an account associated with this GitHub profile.".
1. Verify on prisma studio that there is no change in data.

# Changes:

## Api Logic: `/pages/api/*`

1. Creating user (without old token) is now idempotent (will return 201 if created, 200 if exists already).

## NextJS firebase configs

1. Firebase initialisation will happen outside of login hook for clarity. 
1. The init hook is also renamed from `useAnonymousLoginIfNeeded` to `useFirebaseLogin` as it supports linking.2. 

## Client side api business.

1. `processRequest` in `/api/api.ts` is exported for use in non-standard api calls (those that do not use BaseApi)
1. AxiosInterceptor will no longer modify bearer token if already present (only happens in non-standard api calls)
1. UserApi is a non-standard api call. It will not use BaseApi but instead do `axios.post()` so that it can set bearer token itself.

## Local Storage

1. On every GitHub log in operation, we will save the current JWT to local storage (as an old user token). This is a 1 hour expiry token generated by client side firebase. As it is the same that would have been accessible via browser storage, this is no additional security concern
1. On every log out operation, we will empty the old user token from local storage.
1. If linking account is attempted, we will also empty the old user token from local storage

## Creating new account.

1. On logging in, if no user exists, an anonymous account will be created and added to database via POST /user (no linking).

## Linking GitHub account.

1. When the user opens up the application, if firebase detects that it was due to a redirect operation, it will check redirect results.
1. If redirect results are valid, it will carry on to link the new account to the old account. It retrieves the old token from the local storage and attempts to make it happen. If no old token is found, it will simply attempt to create account.
1. If any error messages are returned, notification alerts will be after logging in (e.g. this account already exists, can't be linked).

# Commits:

- Make api endpoint for creating user idempotent.
- Refactor firebase init out of login hook.
- Rename hook `useAnon...` -> `useFirebaseLogin`.
- Update API & firebase login hook.
- Support interceptor not knowing JWT in advance.
- Use local storage to remember old tokens.
- On login/logout, update local storage.
- Update firebase login to handle redirect flow.
- Add notifications if redirect fails.

### DIFF
--- a/api/api.ts
+++ b/api/api.ts
@@ -41,7 +41,10 @@ class BaseApi {
   }
 }
 
-function processRequest<D extends Payload>(endpoint: string, promise: AxiosPromise<ApiResponse<D>>): ApiPromise<D> {
+export function processRequest<D extends Payload>(
+  endpoint: string,
+  promise: AxiosPromise<ApiResponse<D>>,
+): ApiPromise<D> {
   return promise
     .then((response: AxiosResponse<ApiResponse<D>>) => {
       const apiResponse = response.data;
--- a/api/interceptor.tsx
+++ b/api/interceptor.tsx
@@ -36,7 +36,7 @@ const AxiosInterceptor = ({ children }: Props) => {
         return config;
       }
 
-      config.headers.Authorization = `Bearer ${jwtToken}`;
+      config.headers.Authorization = config.headers.Authorization || `Bearer ${jwtToken}`;
       config.paramsSerializer = toQueryString;
 
       return config;
--- a/api/usersApi.ts
+++ b/api/usersApi.ts
@@ -1,7 +1,7 @@
 import axios from 'axios';
-import { ApiPromise, ApiResponse } from '../types/apiResponse';
-import { UserData, UserPostData } from '../types/user';
-import api from './api';
+import { ApiPromise } from '../types/apiResponse';
+import { UserData } from '../types/user';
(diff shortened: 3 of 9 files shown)


ganhongyao (approved):
Tested and works for both test plans, thanks for the awesome work

sivayogasubramanian (approved):
Tested both and both work.

For the second test, the old UID stays in the database. Is this expected?

zognin:
Screenshot 2022-09-20 at 7 45 22 PM

There is a runtime error when I logged in to github. (Test 1, step 6)

PR #89, opened 2022-09-22
Precache user applications upon login

ganhongyao:


sivayogasubramanian (approved):
Will check in staging

